PR #48, opened 2019-09-03
Allow mocking generic methods with lifetime parameters

asomers:
Now it's possible to mock generic methods with lifetime parameters, as long as they don't also have type parameters, and their return values are `'static` or have the same lifetime as `self`.

@michaelmarconi does this work for you?


michaelmarconi:
@asomers,  that looks like a nicer solution than conditional code compilation!

asomers:
Thanks for your review, and happy hacking!

PR #107, opened 2020-03-12
Fix build when #![deny(missing_docs)] is in force

asomers:
Also, add mor examples to mockall_examples


asomers:
@codehearts everything should build fine with `#[deny(missing_docs)]` now.  Could you please check it?  I tried to add useful docs everywhere.  But there were a few modules that aren't intended for direct user interaction, yet can't be hidden because they contain things that are intended for user interaction.  For those I simply allowed missing docs.

asomers:
The build failure is due to a stricter check for uninitialized data in the latest nightly compiler.  I'll fix it in a separate PR.

codehearts:
This resolved it for me, thank you so much for making time for this!

asomers:
Thanks for reporting it.  And happy hacking!

PR #124, opened 2020-05-03
Fix mem::replace warning

joshleeb:
This PR fixes the following warning caused by not using the returned
value from `mem::replace`.

```
warning: unused return value of `std::mem::replace` that must be used
--> mockall_derive/src/lib.rs:519:21
    |
519 |                     mem::replace(bt, new_bt);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: if you don't need the old value, you can
    just assign the new value directly

warning: 1 warning emitted
```

### DIFF
--- a/mockall_derive/src/lib.rs
+++ b/mockall_derive/src/lib.rs
@@ -77,7 +77,7 @@ cfg_if! {
 
 // If there are any closures in the argument list, turn them into boxed
 // functions
-fn declosurefy(gen: &Generics, args: &Punctuated<FnArg, Token![,]>) -> 
+fn declosurefy(gen: &Generics, args: &Punctuated<FnArg, Token![,]>) ->
     (Generics, Punctuated<FnArg, Token![,]>, Punctuated<TokenStream, Token![,]>)
 {
     let mut hm = HashMap::new();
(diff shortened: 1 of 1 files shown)


mockall_derive/src/lib.rs
```diff
@@ -516,7 +516,7 @@ fn supersuperfy(original: &Type, levels: i32) -> Type {
             Type::BareFn(bfn) => {
                 if let ReturnType::Type(_, ref mut bt) = bfn.output {
                     let new_bt = Box::new(supersuperfy(bt.as_ref(), levels));
-                    mem::replace(bt, new_bt);
+                    let _ = mem::replace(bt, new_bt);
```
joshleeb, line 519: Could instead replace this with `mem::swap` if you prefer :+1: but imho thought this was a bit more explicit

asomers:
Thanks!  But I haven't noticed this warning before.  Is it a new one?  What rustc version are you using?

joshleeb:
> What rustc version are you using?

```
$ rustc --version
rustc 1.45.0-nightly (7ced01a73 2020-04-30)
```

mockall_derive/src/lib.rs
joshleeb: Sorry - should've seen that we could get rid of `mem::replace` entirely

PR #182, opened 2020-08-27
Only run builds through Cirrus if it's a PR or the master branch

asomers:


asomers:
On second thought, this isn't worthwhile.  I could probably make it work, but there are too many corner cases to check.

PR #208, opened 2020-09-07
Fix using QSelf in a where clause or a return type.

asomers:
In where clauses, this is a regression from 0.7.2.  But using qself in
return types never worked correctly.


asomers:
The tests are failing because of an update to the syn crate.  The failure is already fixed on the r0.8.1 branch.  However, I intend to merge this change to the master branch as well.  In order to do that without losing merge history, IIUC, the origin of the feature branch must be the most recent common parent of both r0.8.1 and master.  So I can't rebase this branch.  But I've tested it locally, so I'll merge anyway.

PR #255, opened 2021-02-13
Fix Clippy warnings in generated code with Rust nightly 1.52.0

asomers:


dfreese:
Thanks for your work on this!  Supporting folks that run at `-Dwarnings` with clippy adds work.

Unrelated: being able to use this to fake out tonic server behavior with the addition of async_trait support was a joy in terms of testing.  Thanks!

asomers:
Glad to hear your success story!

PR #261, opened 2021-03-01
corrected transposed letters

Larzenegger:
This seems to be used in the first page that's displayed on docs.rs, and every other usage in examples uses the correct syntax. This minor adjustment should make it less confusing for (other) new Rustaceans.


asomers (approved):
Thanks for your contribution!

Larzenegger:
Do i need to rebase this to benefit from your most recent merge and pass the test?

asomers:
Nope.  We can merge it straight away, without rebasing.

PR #316, opened 2021-07-21
Add count to error message

nbigaouette:
Thanks for the crate! It's quite useful in a project of mine.

The `times()` method on a mock will assert that the expectation has been called a specific number of times.

Unfortunately, it does not print how many times it has been called! Only the expected count...

This makes it harder to diagnose what could have gone wrong in a test.

This PR adds the expectation call count to the panic message. For example:
```
MockFoo::baz: Expectation(<anything>) called 4 times which is more than expected 3
```
or
```
MockFoo::foo: Expectation(<anything>) called 0 time(s) which is fewer than expected 1
```

For this to work, I had to expose a new method `Times::count()`.

I left out any `Cargo.toml` version update as I am not sure of the project's policy on version. Let me know if a version bump is required.

### DIFF
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -1472,7 +1472,11 @@ impl Times {
             if self.range.0.end == 1 {
                 Err("should not have been called".to_owned())
             } else {
-                Err(format!("called more than {} times", self.range.0.end - 1))
+                Err(format!(
+                    "called {} times which is more than expected {}",
+                    count,
+                    self.range.0.end - 1
+                ))
             }
         } else {
             Ok(())
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -1473,7 +1473,7 @@ impl Times {
                 Err("should not have been called".to_owned())
             } else {
                 Err(format!(
-                    "called {} times which is more than expected {}",
+                    "called {} times which is more than the expected {}",
                     count,
                     self.range.0.end - 1
                 ))
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -1483,6 +1483,11 @@ impl Times {
         self.range.0 = 0..usize::max_value();
     }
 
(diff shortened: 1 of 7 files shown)


mockall/src/lib.rs
```diff
@@ -1472,7 +1472,11 @@ impl Times {
             if self.range.0.end == 1 {
                 Err("should not have been called".to_owned())
             } else {
-                Err(format!("called more than {} times", self.range.0.end - 1))
+                Err(format!(
+                    "called {} times which is more than expected {}",
```
asomers, line 1476: I think this would sound more natural:
```suggestion
                    "called {} times which is more than the expected {}",
```

mockall/tests/mock_struct.rs
```diff
@@ -29,52 +29,45 @@ mod checkpoint {
     #[test]
     fn expect_again() {
         let mut mock = MockFoo::new();
-        mock.expect_foo()
-            .returning(|_| 5)
-            .times(1..3);
+        mock.expect_foo().returning(|_| 5).times(1..3);
```
asomers, line 32: Could you please revert the unrelated formatting changes to this file?

asomers (changes requested):
Hey, thanks for the contribution!  It generally looks good.  I just had a few inline comments.  As for the version number, don't worry about it.  I bump it whenever I make a release.  Also, changes to mockall normally require an entry in the CHANGELOG.  But this PR is an exception.  Changes to a human readable error message don't need an entry there.

mockall/tests/mock_struct.rs
nbigaouette: I've re-done commit 65bfd7d as 9af1d7d which should not include format change. Sorry it slipped in...

mockall/src/lib.rs
nbigaouette: I agree! Fixed in d1fffe3.

PR #330, opened 2021-08-30
Fix nondeterministic macro output

dtolnay:
The automock attribute currently creates nondeterministic output, which breaks in environments that cache crate compilations, and causes unnecessary rebuilds of downstream code when the code shouldn't have changed.

Repro:

```rust
#[mockall::automock]
trait Trait {
    fn f<'a, 'b>(s: &'a &'b str);
}
```

Running `cargo expand | rg "for<'., '.>"` this shows mockall sometimes expanding to:

<pre>
Mut(Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> FnMut(&'a &'b str) -&gt; () + Send&gt;),
MutSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> FnMut(&'a &'b str) -&gt; ()&gt;&gt;),
Once(Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> FnOnce(&'a &'b str) -&gt; () + Send&gt;),
OnceSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> FnOnce(&'a &'b str) -&gt; ()&gt;&gt;),
Func(Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> Fn(&&'b str) -&gt; bool + Send&gt;),
FuncSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> Fn(&&'b str) -&gt; bool&gt;&gt;),
Pred(Box&lt;(Box&lt;dyn <b>for&lt;'b, 'a&gt;</b> ::mockall::Predicate&lt;&'b str&gt; + Send&gt;,)&gt;),
fn with&lt;MockallMatcher0: <b>for&lt;'b, 'a&gt;</b> ::mockall::Predicate&lt;&'b str&gt; + Send + 'static&gt;(
</pre>

and sometimes to:

<pre>
Mut(Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> FnMut(&'a &'b str) -&gt; () + Send&gt;),
MutSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> FnMut(&'a &'b str) -&gt; ()&gt;&gt;),
Once(Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> FnOnce(&'a &'b str) -&gt; () + Send&gt;),
OnceSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> FnOnce(&'a &'b str) -&gt; ()&gt;&gt;),
Func(Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> Fn(&&'b str) -&gt; bool + Send&gt;),
FuncSt(::mockall::Fragile&lt;Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> Fn(&&'b str) -&gt; bool&gt;&gt;),
Pred(Box&lt;(Box&lt;dyn <b>for&lt;'a, 'b&gt;</b> ::mockall::Predicate&lt;&'b str&gt; + Send&gt;,)&gt;),
fn with&lt;MockallMatcher0: <b>for&lt;'a, 'b&gt;</b> ::mockall::Predicate&lt;&'b str&gt; + Send + 'static&gt;(
</pre>

The source of the nondeterminism is iterating over a randomized HashSet.

This PR fixes the nondeterminism and uses https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_type to disallow iteration of hash based collections in the future.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,2 +1,9 @@
 # I use a lot of dumb names in the tests
 blacklisted-names = []
+
+# It is naughty for a code generator to produce nondeterministic output.
+# Use UnorderedMap/Set instead.
+disallowed-types = [
+    "std::collections::HashMap",
+    "std::collections::HashSet",
+]
--- a/mockall_derive/src/automock.rs
+++ b/mockall_derive/src/automock.rs
@@ -1,6 +1,6 @@
 // vim: tw=80
 use super::*;
-use std::collections::HashMap;
+use crate::collections::UnorderedMap;
 use syn::parse::{Parse, ParseStream};
 
 /// A single automock attribute
@@ -27,7 +27,7 @@ impl Parse for Attr {
 /// automock attributes
 #[derive(Debug, Default)]
 pub(crate) struct Attrs {
-    pub attrs: HashMap<Ident, Type>,
+    pub attrs: UnorderedMap<Ident, Type>,
     pub modname: Option<Ident>
 }
 
@@ -249,7 +249,7 @@ impl Attrs {
 
 impl Parse for Attrs {
     fn parse(input: ParseStream) -> parse::Result<Self> {
-        let mut attrs = HashMap::new();
+        let mut attrs = UnorderedMap::new();
         let mut modname = None;
         while !input.is_empty() {
             let attr: Attr = input.parse()?;
--- a/mockall_derive/src/collections.rs
+++ b/mockall_derive/src/collections.rs
@@ -17,21 +17,23 @@ use std::hash::Hash;
 use std::iter::FromIterator;
 
 #[derive(Debug)]
-pub struct UnorderedMap<K, V>(HashMap<K, V>);
+pub struct UnorderedMap<K, V>(HashMap<K, V>)
+where
+    K: Eq + Hash;
 
 #[derive(Debug)]
-pub struct UnorderedSet<T>(HashSet<T>);
-
-impl<K, V> UnorderedMap<K, V> {
-    pub fn new() -> Self {
-        UnorderedMap(HashMap::new())
-    }
-}
+pub struct UnorderedSet<T>(HashSet<T>)
+where
+    T: Eq + Hash;
 
 impl<K, V> UnorderedMap<K, V>
 where
     K: Eq + Hash,
 {
+    pub fn new() -> Self {
+        UnorderedMap(HashMap::new())
+    }
+
     pub fn insert(&mut self, k: K, v: V) -> Option<V> {
         self.0.insert(k, v)
     }
@@ -74,13 +76,19 @@ where
     }
 }
 
-impl<K, V> Default for UnorderedMap<K, V> {
+impl<K, V> Default for UnorderedMap<K, V>
+where
+    K: Eq + Hash,
+{
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<T> Default for UnorderedSet<T> {
+impl<T> Default for UnorderedSet<T>
+where
+    T: Eq + Hash,
+{
     fn default() -> Self {
         UnorderedSet(HashSet::default())
     }
--- a/mockall_derive/src/collections.rs
+++ b/mockall_derive/src/collections.rs
@@ -1,0 +1,99 @@
+//! A map and set that are safe for code generator use.
+//!
+//! *Do not add any IntoIterator impl to these.* Only insertion, membership
+//! checking, and map lookup are exposed. Iteration order is not exposed.
+//!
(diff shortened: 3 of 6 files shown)


asomers:
Interesting find!  I promise to review this by the weekend, but I may not have time until then.

asomers (changes requested):
While the problem is real, your solution seems excessively complicated.  What's wrong with just using a deterministic HashMap/HashSet?  They're only random because they use a random hasher by default.  But they can be made deterministic by constructing one like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=420f28eeccedee9cd077fff59f831635 .

Also, this bug needs a regression test.  I'll work on that.

asomers:
If you check the "allow edits by maintainers" button, I'll push the test to the PR.

dtolnay:
@asomers, I believe the complexity over your proposed alternative is justified because that's what makes this mistake machine catchable. Replacing all current places that create a randomized hashmap with `with_hasher` is not sufficient because every future code change would need to be inspected for this bug, and at least for myself as a reviewer I don't have confidence that I could keep all the issues requiring vigilance in my head.

asomers:
That's what the test is for.  Also, would your `disallowed-types` trick work for `RandomState`, as used indirectly by `HashMap::default()`?

dtolnay:
> That's what the test is for.

A test can (probabilistically) catch nondeterminism in a handful of automock invocations -- the ones covered by the test. Whereas banning HashMap rules out the entire category of HashMap-related nondeterminism across all possible automock invocations, including ones not well covered by the test suite.

I do agree with you that another test would be good, because a test could catch new sources of nondeterminism outside of HashMap. In practice that's less of a concern to me because I've debugged many of these nondeterministic compilation issues and it's been HashMap iteration **every time**, with the one exception of https://crates.io/crates/const-random which is blatant and intentional.

> Also, would your `disallowed-types` trick work for `RandomState`, as used indirectly by `HashMap::default()`?

It does not catch this as far as I can tell. The mechanism here is the same as what is used for deprecations -- it's catching whether your code has *named* the thing. If the only place the word `RandomState` is written is in std code somewhere, `disallowed-types` would consider that std's fault, and not flag it in the downstream code which does not name `RandomState`.

asomers:
Well, `HashMap<K, V>` is really just shorthand for `HashMap<K, V, RandomState>`.  Is clippy smart enough to detect that?  Or can it detect method calls like `HashMap::new` and `HashMap::default`?  I'm just concerned about the size of your patch.  I don't want to maintain a whole new collection type if I don't have to.

dtolnay:
> Well, `HashMap<K, V>` is really just shorthand for `HashMap<K, V, RandomState>`. Is clippy smart enough to detect that?

I think this isn't how the linter works. (But you could propose an enhancement to clippy. I don't know enough about the linter infrastructure to know how reasonable this would be to implement.)

`HashMap<K, V>` does not literally mention the type `RandomState`. It's the same for deprecation:

```rust
use std::marker::PhantomData;

#[deprecated]
pub struct Thing;

#[allow(deprecated)]
pub struct Generic<T = Thing>(PhantomData<T>);

fn main() {
    // does not mention "Thing", no warning
    let _g: Generic = Generic(PhantomData);
}
```

> Or can it detect method calls like `HashMap::new` and `HashMap::default`?

There is a `disallowed_method` lint that you could use to flag `HashMap::new`. It can't flag `HashMap::default` unless you want to ban all of `std::default::Default::default`. But even beyond those, there are infinite ways to obtain a HashMap (e.g. think of deserializing it from JSON. The HashMap *construction* appears nowhere in your code.) I would have little confidence that a new one wouldn't sneak in over time.

> I don't want to maintain a whole new collection type if I don't have to.

I think the reality is there isn't an appropriate collection type in std, so for a solution that involves banning iteration in collection order, you would either need it in this crate or you would need to add a dependency that provides it (which is an option if you are open to it). FWIW it's adding \<1% to the Rust code already in this crate.

asomers:
@dtolnay I ended up going the "HashMap with BuildHasherDefault" route.  But I did make one new change: instead of a single test case for nondeterministic output, I added a `--cfg` option which checks for nondeterminism on every single invocation.  Then I added a CI test step that builds the entire test suite with that option.  I think that should pretty well cover it.  Thanks again for raising this issue.

Superseded by #333 .

PR #337, opened 2021-09-20
docs: update docs with synchronization example

TomPridham:
just adds a link to the synchronization example in the main lib.rs doc

### DIFF
--- a/mockall/examples/synchronization.rs
+++ b/mockall/examples/synchronization.rs
@@ -1,0 +1,75 @@
+// vim: tw=80
+//! Add synchronization to multiple tests that are accessing the same mock
+//!
+//! When mockall mocks a function or static method, it does so globally. This
+//! can cause hard to debug and non-deterministic failures when one test
+//! overwrites the mock that another test is depending on. The solution to this
+//! is to add some form of synchronization so that tests that depend on a
+//! specific mock will not run in parallel. This is easily achieved using a
+//! Mutex and the lazy_static crate.
+#![deny(warnings)]
+
+use mockall_double::double;
+
+pub mod my_mock {
+    #[cfg(test)]
+    use mockall::automock;
+
+    pub struct Thing;
+    #[cfg_attr(test, automock)]
+    impl Thing {
+        pub fn one() -> u32 {
+            1
+        }
+    }
+}
+
+#[double]
+use my_mock::Thing;
+
+fn main() {
+    println!("1 == {}", Thing::one());
+}
+
+#[cfg(test)]
+mod test {
+    use crate::my_mock::MockThing;
+    use lazy_static::lazy_static;
+    use std::sync::{Mutex, MutexGuard};
+
+    lazy_static! {
+        static ref MTX: Mutex<()> = Mutex::new(());
+    }
+
+    // When a test panics, it will poison the Mutex. Since we don't actually
+    // care about the state of the data we ignore that it is poisoned and grab
+    // the lock regardless.  If you just do `let _m = &MTX.lock().unwrap()`, one
+    // test panicking will cause all other tests that try and acquire a lock on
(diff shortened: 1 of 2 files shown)


asomers:
This is probably a good addition, once you fix rustdoc.  Even better though would be dedicated synchronization examples, perhaps in the examples directory.
Also, I've considered building synchronization into Mockall.  Basically, creating a Context object would automatically acquire a lock.  However, these would be vulnerable to lock-order reversals and other bugs, since the user would probably be unaware of them.  What do you think of that idea?

mockall/src/lib.rs
```diff
@@ -1132,11 +1132,10 @@ use std::{
     any,
     fmt::{self, Debug, Formatter},
     marker::PhantomData,
-    ops::{Range, RangeFrom, RangeFull, RangeInclusive, RangeTo,
```
TomPridham, line 1135: i have rustfmt run on save. i can revert the formatting changes if you care

TomPridham:
it seems like the synchronization needs to happen between all modules that are using the same mock. i'll update the example tomorrow to better illustrate that. i just did some refactoring that exposed that potential failure mode in our own codebase :sweat_smile:

asomers:
> does the synchronization need to happen between all different test mods or are the different test modules isolated? i have two files with test modules `a.rs` and `b.rs` that both mock `c.rs`. does the synchronization need to make sure that only one test is running at a time between `a` AND `b` or is it enough to make sure that only one test at a time is running per module?

It depends on how the two modules are linked.  If they're in the same process, then you need synchronization between them.  If they are separate processes, then you don't.  By default, Cargo will create a separate program for every file in the tests directory, and all tests within that program get executed as part of the same process.

mockall/examples/synchronization.rs
```diff
@@ -0,0 +1,73 @@
+// vim: tw=80
+//! Add synchronization to multiple tests that are accessing the same mock
+//!
+//! When mockall mocks a function or static method, it does so globally. This can cause hard to debug and
```
asomers, line 4: Please wrap to 80 cols.

TomPridham: sorry it took so long,but that's done

mockall/src/lib.rs
```diff
@@ -836,7 +836,8 @@
 //!
 //! Mockall can also mock static methods.  But be careful!  The expectations are
 //! global.  If you want to use a static method in multiple tests, you must
-//! provide your own synchronization.  For ordinary methods, expectations are
+//! provide your own synchronization. See the [`synchronization example`](examples/synchronization.rs)
```
asomers, line 839: This link will be dead, because rustdoc doesn't include files in the examples/ directory.  Instead, it's probably best to link to the file on Github.

asomers (changes requested):

mockall/src/lib.rs
TomPridham: i just linked to where the file would be once this is merged in. should i link to a specific commit hash or is that fine?

asomers: I think it's best not to provide a specific hash.  We might update the example later, and it would be easy to forget to update the link.

TomPridham: okay, updated to use my previous commit as the reference

asomers: It looks like you've pinned a specific commit.  Don't you think it's better to point to the latest version?

TomPridham: oh, oops. i misread your comment. i thought you wanted me to include a specific commit hash. i can revert that to point to the latest version

TomPridham: @asomers does this seem good to go now?

asomers (approved):
Yep, it looks good now.

PR #364, opened 2022-02-25
fix `Box<dyn A>: A` required in returning `&dyn A` for #363

### LINKED ISSUE #363: trait `A` is not implemented for `Box<dyn A>`
Hi folks, I met a compile error when doing this:
```rust
trait A { }

trait B: A {
    fn as_a(&self) -> &dyn A;
}


mockall::mock! {
    Foo {}
    impl A for Foo {}
    impl B for Foo {
        fn as_a(&self) -> &dyn A;
    }
}
```
The compile error is:
```
error[E0277]: the trait bound `Box<dyn A>: A` is not satisfied
  --> src/lib.rs:11:1
   |
11 | / mockall::mock! {
12 | |     Foo {}
13 | |     impl A for Foo {}
14 | |     impl B for Foo {
15 | |         fn as_a(&self) -> &dyn A;
16 | |     }
17 | | }
   | |_^ the trait `A` is not implemented for `Box<dyn A>`
   |
   = note: required for the cast to the object type `dyn A`
   = note: this error originates in the macro `mockall::mock` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `mocktest` due to previous error
```
I ran `cargo expand` to see what was happening, and find that:
```rust
impl __mock_MockFoo_B::__as_a::Expectation {
    /// Call this [`Expectation`] as if it were the real method.
    pub(in super::super) fn call(&self) -> &Box<dyn A> {
         /* ... */
    }
}
// ...
impl B for MockFoo {
    fn as_a(&self) -> &dyn A {
        let no_match_msg = /* something irrelavent ... */;
        self.B_expectations.as_a.call().expect(&no_match_msg)
    }
}
```
Here `Expectation::call` returns `&Box<dyn A>` but `B::as_a` requires `&dyn A`, and `Box<dyn A>: A` is not satisfied.

The solution is simple. I just add something like this:
```rust
impl<T: A + ?Sized> A for Box<T> {}
```
Then it compiles.

---
But is it possible to make this change in the generated code in `MockFoo::as_a`?
```diff
-        self.B_expectations.as_a.call().expect(&no_match_msg)
+        &*self.B_expectations.as_a.call().expect(&no_match_msg)
```
After that, there is no extra `impl<T: A + ?Sized> A for Box<T> {}` needed.

frank-king:
fixes #363.

### DIFF
--- a/mockall/tests/mock_return_dyn_trait.rs
+++ b/mockall/tests/mock_return_dyn_trait.rs
@@ -10,6 +10,23 @@ trait T: Debug + Sync {
 }
 
 impl T for u32 {}
+trait Base {
+    fn as_t(&self) -> &dyn T;
+}
+trait Derived: Base {
+    fn as_base(&self) -> &dyn Base;
+}
+
+impl Base for u32 {
+    fn as_t(&self) -> &dyn T {
+        self
+    }
+}
+impl Derived for u32 {
+    fn as_base(&self) -> &dyn Base {
+        self
+    }
+}
 
 impl<Q> T for Q where Q: Debug + Sync + AsMut<dyn T> {}
 
@@ -18,6 +35,16 @@ mock!{
         fn foo(&self) -> &dyn Debug;
         fn bar(&self) -> &'static dyn T;
         fn baz(&mut self) -> &mut dyn T;
+    }
+}
+
+mock! {
+    Bar {}
+    impl Base for Bar {
+        fn as_t(&self) -> &dyn T;
+    }
+    impl Derived for Bar {
+        fn as_base(&self) -> &dyn Base;
     }
 }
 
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
This looks almost good.  In addition to the inline comments, it also needs a CHANGELOG entry.  I suggest
```

### Fixed

- Fixed mocking methods that return a reference to a `dyn T` trait object,
  when that trait isn't already implemented for `Box<dyn t>`.
  ([#364](https://github.com/asomers/mockall/pull/364))
```

onalante-msft:
If this PR is awaiting the fixes, I can commit the recommended changes. Without this fix, `mockall` cannot be used on---among other things---OpenAPI-generated clients[^0] that would be very helpful to mock.

[^0]: Example: https://github.com/OpenAPITools/openapi-generator/blob/6.0.x/modules/openapi-generator/src/main/resources/rust/hyper/client.mustache#L12

asomers:
Sure @onalante-msft  that would be helpful.  Also, I think a rebase will fix the test.

asomers:
superseded by #380 .

PR #375, opened 2022-04-03
feat: create an argument captor

### LINKED ISSUE #366: capture argument values for further assertions
When mocking a method that takes complex structures as parameters, it is common to prime the mock to capture that argument, and subsequently assert on it. See e.g. https://site.mockito.org/javadoc/current/org/mockito/ArgumentCaptor.html.

I was able to achieve the same with little code:
```rust
    pub struct Captor<T: Clone + 'static>(Arc<Mutex<Option<T>>>);

    pub fn captor<T: Clone + 'static>() -> Captor<T> {
        Captor(Arc::new(Mutex::new(None)))
    }

    impl<T: Clone + 'static> Captor<T> {
        pub fn capture(&self) -> impl Fn(&T) -> bool {
            let store = self.0.clone();
            move |value| {
                *store.lock().unwrap() = Some(value.clone());
                true
            }
        }

        pub fn value(self) -> Option<T> {
            self.0.lock().unwrap().take()
        }
    }
```

(maybe it would be more flexible to use `Vec<T>` instead of `Option<T>`).

However, I believe that this is basic functionality, expected to be provided by a mocking library. `Captor` should just be another `Predicate`.

Thanks.

albx79:
Fixes #366

### DIFF
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -1694,31 +1694,36 @@ pub struct Captor<T> {
     values: Arc<Mutex<Vec<T>>>,
 }
 
+/// Create a new captor. See the documentation for [Captor](Captor)
 pub fn captor<T>() -> Captor<T> {
     Captor { values: Arc::new(Mutex::new(vec![])) }
 }
 
 impl<T: Clone> Captor<T> {
+    /// Retrieve the list of all captured values, in chronological order
     pub fn values(&self) -> Vec<T> {
         self.lock().clone()
     }
 
+    /// Retrieve the last captured value, if any.
     pub fn last_value(&self) -> Option<T> {
         self.lock().last().cloned()
     }
 
+    /// Forget all previously captured values. Reset this `Captor` to its initial state.
     pub fn reset(&self) {
         self.lock().clear()
     }
 
-    fn lock(&self) -> MutexGuard<Vec<T>> {
-        self.values.lock().unwrap()
-    }
-
+    /// Start capturing values. See the documentation for [Captor](Captor)
     pub fn capture_values(&self) -> impl Predicate<T> {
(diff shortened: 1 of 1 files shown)


albx79:
I wasn't sure where to put the new code, so I put it at the end of `lib.rs`. Please let me know if it should go somewhere else.

The example in the rustdoc effectively acts as a test suite (it exercises all code paths).

mockall/src/lib.rs
```diff
@@ -1656,3 +1659,85 @@ impl Sequence {
...
+        self.lock().last().cloned()
+    }
+
+    /// Forget all previously captured values. Reset this `Captor` to its initial state.
+    pub fn reset(&self) {
+        self.lock().clear()
+    }
+
+    /// Start capturing values. See the documentation for [Captor](Captor)
+    pub fn capture_values(&self) -> impl Predicate<T> {
```
asomers, line 1719: Just in the interest of reducing typing, I suggest shortening the name:
```suggestion
    pub fn capture(&self) -> impl Predicate<T> {
```

mockall/src/lib.rs
```diff
@@ -1656,3 +1659,85 @@ impl Sequence {
...
+
+    fn lock(&self) -> MutexGuard<Vec<T>> {
+        self.values.lock().unwrap()
+    }
+}
+
+impl<T: Clone> Predicate<T> for Captor<T> {
+    fn eval(&self, variable: &T) -> bool {
+        self.values.lock().unwrap().push(variable.clone());
+        true
```
asomers, line 1733: Instead of always evaluating true, would it be possible to chain `Captor` with another predicate?

asomers (changes requested):

PR #389, opened 2022-06-28
Fix using Mockall when a function named `Ok` is in scope

### LINKED ISSUE #388: use-ing anyhow::Ok interferes with generated code
I'm filing the issue mostly because this results in a somewhat non-obvious error message.

Repro:

```Rust
use anyhow::{Error, Ok};
use async_trait::async_trait;
use mockall::automock;

#[automock]
#[async_trait]
pub trait Foo {
    async fn bar<'a>(&self) -> Result<(), Error>;
}
```

Compiler error:
```
54 | #[automock]
   | ^^^^^^^^^^^
   | |
   | expected `&str`, found struct `anyhow::Error`
   | expected `Result<Result<(), anyhow::Error>, &'static str>` because of return type
   |
   = note: expected enum `Result<_, &'static str>`
              found enum `Result<_, anyhow::Error>`
   = note: this error originates in the attribute macro `automock` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Changing the first line to `use anyhow::Error;` causes the issue to go away.

It seems like some of the generated code (like `fn call_mut(&mut self) -> std::result::Result<Result<(), Error>, &'static str>`) explicitly references `std::result::Result` while other bits (like returning `Ok(__mockall_f())`) rely on nobody being silly enough to import their own `Ok`.
Hopefully fixing this is as simple as having the generated code return a fully-qualified `std::result::Result::Ok` instead of just `Ok`.

asomers:
The `anyhow` create defines a function by this name, and the proc-macro
generated code was picking it up instead of the standard Result::Ok .

Fixes #388

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 
+## [ Unreleased ] - ReleaseDate
+
+## Fixed
+
+- Fixed using Mockall when a function named `Ok` is in scope.  The `anyhow`
+  crate, for example, creates a function by this name.
+  ([#389](https://github.com/asomers/mockall/pull/389))
+
 ## [ 0.11.1 ] - 2022-05-15
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


asomers:
@mkosiba can you test this PR with your project?

PR #390, opened 2022-06-28
docs: add some docs about using mock! to mock clone derives

TomPridham:
i was questioning how to mock `#[derive(Clone)]` and saw there were a few issues with people wondering the same thing. hopefully this clears up future questions


asomers (approved):
Thanks.  This looks like a nice example.

PR #416, opened 2022-10-11
add static method caveat to Modules documentation

bryanlarsen:
If this warning was in the documentation it would have saved me a few hours.


asomers:
And what caveats are those ?  The part about needing to use the `foo_context` method or something?

asomers (approved):
Allright, I think this is a good addition.

PR #421, opened 2022-10-23
Allow mocking functions with unknown size type bounds

HampusMat:
A quick change to allow for mocked functions to have type bounds with unknown sizes.

## For example:
```rust
use mockall::mock;

trait Foo {}

trait Bar {
    fn do_something<T>(&self) -> u32
    where
        T: 'static + ?Sized;
}

mock! {
    Foobar {}

    impl Bar for Foobar {
        fn do_something<T>(&self) -> u32
        where
            T: 'static + ?Sized;
    }
}

let mut foobar_mock = MockFoobar::new();

foobar_mock
    .expect_do_something::<dyn Foo>()
    .returning(|| 120);

let num = foobar_mock.do_something::<dyn Foo>();

println!("{num}");
```


mockall/tests/automock_generic_method_with_unknown_size_bounds.rs
```diff
@@ -0,0 +1,38 @@
+// vim: tw=80
+//! generic methods with bounds on their generic parameters
```
asomers, line 2: This comment looks copy-pasted.  Could you please update it?

mockall/tests/automock_generic_method_with_unknown_size_bounds.rs
```diff
@@ -0,0 +1,38 @@
...
+#![deny(warnings)]
+
+use mockall::*;
+use std::fmt::Debug;
+
+struct X<T: Debug + ?Sized>(Box<T>);
```
asomers, line 8: This test program is overly complicated.  You can eliminate all references to `X` and `Debug` and just test with a method signature like this:
```rust
#[automock]
trait Foo {
    fn foo<T: 'static + ?Sized>(&self, x: Box<T>);
}
```

mockall/tests/mock_generic_method_with_unknown_size_type_bound.rs
```diff
@@ -0,0 +1,76 @@
+// vim: tw=80
```
asomers, line 1: It looks like you copied this test program from mock_generic_method_with_where_clause.rs.  In that case, it really was necessary to have the four different methods, because different code gets emitted in each case.  But here, having the four separate methods doesn't give you any better coverage.  In fact, even testing mock and automock separately doesn't improve the code coverage of this feature.  I think you can eliminate this file altogether.

asomers (changes requested):
This is a nice little fix.  Don't forget to add it to the "Fixed" section of the CHANGELOG!

asomers (approved):

PR #425, opened 2022-11-11
Print better expectation messages on stable

jefftt:
This change allows better messages for "No matching expectation found" on stable. This is a great user experience upgrade as it allows this behavior without requiring running with the nightly toolchain. This MR applies the generalized technique for autoderef specialization described [here](https://lukaskalbertodt.github.io/2019/12/05/generalized-autoref-based-specialization.html)


jefftt:
Looks like it does regress on generic arguments though: `tests/mock_generic_arguments.rs`. I think we can use this on stable and use the current specialization implementation on nightly?

mockall_derive/src/mock_function.rs
```diff
@@ -595,9 +599,9 @@ impl MockFunction {
         } else {
             format!("{}::{}", self.mod_ident, self.sig.ident)
         };
-        let fields = vec!["{:?}"; argnames.len()].join(", ");
+        let fields = vec!["{}"; argnames.len()].join(", ");
         let fstr = format!("{}({})", name, fields);
-        quote!(std::format!(#fstr, #(::mockall::MaybeDebugger(&#argnames)),*))
+        quote!(std::format!(#fstr, #((&&&::mockall::MaybeDebugger(&#argnames)).debug_string()),*))
```
asomers, line 604: `debug_string()` will allocate space for a string and populate it.  Then `std::format!` will do the same.  That's less efficient than the original.  Could we change the autoderef magic to create an actual `Debug` implementation like before, rather than use this `debug_string` method?

mockall/tests/mock_generic_arguments.rs
```diff
@@ -6,7 +6,7 @@ use mockall::*;
 
 mock! {
     Foo {
-        fn foo<T: 'static>(&self, t: T) -> i32;
+        fn foo<T: 'static + std::fmt::Debug>(&self, t: T) -> i32;
```
asomers, line 9: Any reason not to import Debug in the usual way, with `use std::fmt::Debug`?

mockall/tests/mock_generic_struct_with_generic_static_method.rs
```diff
@@ -6,12 +6,12 @@ use mockall::*;
 use std::sync::Mutex;
 
 mock! {
-    Foo<T: 'static> {
-        fn foo<Q: 'static>(t: T, q: Q) -> u64;
+    Foo<T: 'static + std::fmt::Debug> {
```
asomers, line 9: Why do you need to add `Debug`?  Shouldn't the autoderef magic pretty-print the panic message even without this trait bound?

asomers (changes requested):
Sorry it took me so long to review this.  It's just been hard to find a solid block of time.  I must say, this autoderef stuff is a big hack.  But such a useful hack!  It's so very useful, I don't think that Mockall can do without it.  My biggest question is, "why bother with `Display`"?  I think that `Debug` is more appropriate for this usage.  If a type implements both, I think that `Debug` should get higher priority.  And types that implement `Display` but not `Debug` are rarer than hens' teeth, so I don't think we should worry about them.  Can we just eliminate the `Display` part?
Also, it needs a CHANGELOG entry.

mockall_derive/src/mock_function.rs
jefftt: played around with this some more and got something that wont allocate unnecessarily!

mockall/tests/mock_generic_struct_with_generic_static_method.rs
jefftt: unfortunately not, removing the bounds falls back to the `"?"` message. not sure if i'm missing anything else to make it work

jefftt:
> I must say, this autoderef stuff is a big hack. But such a useful hack! It's so very useful

I agree :) it's definitely an abuse of the type system, but very nice to not have to depend on nightly

> My biggest question is, "why bother with `Display`"? I think that `Debug` is more appropriate for this usage. If a type implements both, I think that `Debug` should get higher priority. And types that implement `Display` but not `Debug` are rarer than hens' teeth, so I don't think we should worry about them. Can we just eliminate the `Display` part? 

This was based on my very narrow use case of printing [Date](https://docs.rs/time/latest/time/struct.Date.html) which, at the time, had a much better display implementation. Looks like they recently improved this so my point is now moot. https://github.com/time-rs/time/pull/509. Removed the display trait

asomers (approved):
This looks great!  Would you just mind rebasing to fix the conflict?

jefftt:
sorry for the late reply, i've rebased the branch on latest master

asomers (changes requested):
Could you please either revert the pure-formatting changes, or else move them into a separate commit?

asomers (approved):

jefftt:
@asomers hey sorry to bother you, but do you plan on including this in a release in the future? would like to use it in my projects without having to do a git override

asomers:
Yes, I will include it in the next feature release.  I haven't published one yet because I'm still not completely happy with the concretize feature.  See https://users.rust-lang.org/t/how-to-pass-multiple-items-to-a-derive-macro/91266 for more discussion about that.

PR #429, opened 2022-11-20
Fix link to associated types

saikatdas0790:


asomers (approved):
Whoops!  Thanks for the fix.

PR #443, opened 2022-12-23
Always drop static methods' Expectations during Drop

### LINKED ISSUE #442: static functions' expectations don't get cleared after a panic.
I have an issue using mocks for static methods.
I was able to reproduce it with a slight modification of the tests (https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs):
1. Replace `returning` with `return_once`.
2. Change the expectation in the first test to fail it.
3. Now the second test starts to fail with `MockThing::one: Expectation(<anything>) called twice, but it returns by move`

asomers:
Even if a panic is in-progress.  Originally we didn't drop them during panic, for fear of double-panicing.  But there is already protection from double-panic in Common::drop.  Not clearing expectations during a panic can cause surprising failures in one test case after a different test case panics.

Fixes #442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Static methods' expectations will now be cleared during a panic.
+  ([#443](https://github.com/asomers/mockall/pull/443))
+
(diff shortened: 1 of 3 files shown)


asomers:
@DGolubets does this fix your problem?

DGolubets:
@asomers Yes, it does. Thanks a lot!

PR #447, opened 2023-01-16
Change Getting Started to a code that compiles

mfdutra:
The new code snippet can be copied/pasted and it will run as is with cargo test.

For someone that is starting with testing/mocking in Rust, I think it would be nice to get a code snipped that just works out of the box.

### DIFF
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -58,16 +58,24 @@
 //!     fn foo(&self, x: u32) -> u32;
 //! }
 //!
-//! fn call_with_four(x: &MyTrait) -> u32 {
+//! fn call_with_four(x: &impl MyTrait) -> u32 {
 //!     x.foo(4)
 //! }
 //!
-//! let mut mock = MockMyTrait::new();
-//! mock.expect_foo()
-//!     .with(predicate::eq(4))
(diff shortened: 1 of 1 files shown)


asomers:
Were you confused by the fact that doc tests have an implied `main` function?

mfdutra:
Actually no. What caught my attention was the invalid signature for `call_with_four`, that is missing `&impl`.

I added the boilerplate around the test code just to make it runnable.

mfdutra:
```
trait MyTrait {
    fn foo(&self, x: u32) -> u32;
}

fn call_with_four(x: &MyTrait) -> u32 {
    x.foo(4)
}
```

That will not compile, unless:

```
fn call_with_four(x: &impl MyTrait) -> u32 { 
...
```

...or something equivalent.

asomers:
Oh, that's an edition thing.  It works fine in edition 2018, which is what Mockall uses.

asomers:
I'll separately raise the edition to 2021, because the MSRV is already high enough.  But I don't want the other parts of this PR.  I think they just make the examples more confusing.

mfdutra:
I've updated my branch just adding `&impl` then, in case you want to pull.

PR #472, opened 2023-03-21
Check if actual calls exceed the expected number on mock object drop

vikz95:
This pull request will allow to check again if the actual number of function calls is greater than the expected during the mock object dropout.
So if the check executed at every call panics in a sub task, this additional check at object dropout will cause another panic and will make the test fail. 

I was not able to run `cargo test`  because I got this error:

```
error[E0601]: `main` function not found in crate `ffi`
  --> mockall\examples\ffi.rs:35:2
   |
35 | }
   |  ^ consider adding a `main` function to `mockall\examples\ffi.rs`
```


mockall/src/lib.rs
```diff
@@ -1605,7 +1605,8 @@ impl Times {
     /// Has this expectation already been called the minimum required number of
```
asomers, line 1605: You should update doc comment too, because you've changed the function's behavior.

mockall_derive/src/mock_function.rs
```diff
@@ -416,9 +416,9 @@ pub(crate) struct MockFunction {
     /// Types used for Predicates.  Will be almost the same as args, but every
     /// type will be a non-reference type.
     predty: Vec<Type>,
-    /// Does the function return a non-'static reference? 
+    /// Does the function return a non-'static reference?
```
asomers, line 419: Please do not mix formatting changes and content changes in the same commit.

mockall_derive/src/mock_function.rs
```diff
@@ -988,13 +988,17 @@ impl<'a> ToTokens for Common<'a> {
...
                     if !::std::thread::panicking() && !self.times.is_satisfied()
                     {
-                        let desc = std::format!(
-                            "{}", self.matcher.lock().unwrap());
-                        panic!("{}: Expectation({}) called {} time(s) which is fewer than expected {}",
-                               #funcname,
-                               desc,
-                               self.times.count(),
-                               self.times.minimum());
+                        if (self.times.count() < self.times.minimum()) {
```
asomers, line 991: This is very confusing because you're checking times.count() here, and also implicitly up above in is_satisfied.  It makes the code hard to read.  You should only do that check on one level.

asomers (changes requested):
This needs a CHANGELOG and a test case too.

vikz95:
Created a new pull request for the test failure on Windows: https://github.com/asomers/mockall/pull/473

I'll fix the other issues tomorrow.

mockall_derive/src/mock_function.rs
vikz95: Could you elaborate on this point?

Now is_satisfied checks both expected boundaries of the actual number of calls and returns a boolean.
From a boolean I'm not able to determine what error message to print in the panic, this is why the second check.

Instead of a boolean I could return an Enum with values: SATISFIED_LOWER_BOUND, SATISFIED_UPPER_BOUND, SATISFIED, NOT_SATISFIED. 
Then I could differentiate the cases. But I don't know if this is what you had in mind...

asomers: You could do something like that, returning an enum.  Or you could inline the is_satisfied logic into the proc macro.  Either could work.  I just don't want that logic to be repeated in two places.

mockall_derive/src/mock_function.rs
```diff
@@ -986,15 +986,22 @@ impl<'a> ToTokens for Common<'a> {
...
-                    {
-                        let desc = std::format!(
-                            "{}", self.matcher.lock().unwrap());
-                        panic!("{}: Expectation({}) called {} time(s) which is fewer than expected {}",
-                               #funcname,
-                               desc,
-                               self.times.count(),
-                               self.times.minimum());
+                    if !::std::thread::panicking() {
+                        match self.times.is_satisfied(); {
```
vikz95, line 990: will remove ; in next commit

Why cargo build doesn't catch there errors?
How should I compile?

vikz95:
If the implementation with enums is ok, then I will create a test where the panics are raised in a subthread and then assure that checkpoint() makes the test fail.

I also need to figure why now cargo test causes so many errors, maybe it's due to the new enum:
```
error[E0433]: failed to resolve: use of undeclared type `ExpectedCalls`                                                                                                                
 --> mockall\tests\automock_attrs.rs:7:1
  |
7 | #[automock]
  | ^^^^^^^^^^^ use of undeclared type `ExpectedCalls`
  |

```

mockall_derive/src/mock_function.rs
asomers: Well, if you just did `cargo build`, that would build the proc macros and the small amount of code in the mockall crate, but it wouldn't evaluate any of the proc macros, anywhere.  You should do `cargo test` before pushing.

asomers:
> If the implementation with enums is ok, then I will create a test where the panics are raised in a subthread and then assure that checkpoint() makes the test fail.

No need to get all fancy with threads.  Just use [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html#).

> 
> I also need to figure why now cargo test causes so many errors, maybe it's due to the new enum:
> 
> ```
> error[E0433]: failed to resolve: use of undeclared type `ExpectedCalls`                                                                                                                
>  --> mockall\tests\automock_attrs.rs:7:1
>   |
> 7 | #[automock]
>   | ^^^^^^^^^^^ use of undeclared type `ExpectedCalls`
>   |
> ```

Because you forgot to import that type in the generated code.  Best to refer to it by its absolute path, `::mockall::ExpectedCalls`

mockall_derive/src/mock_function.rs
```diff
@@ -986,15 +986,22 @@ impl<'a> ToTokens for Common<'a> {
...
+                                let desc = std::format!(
+                                    "{}", self.matcher.lock().unwrap());
+                                panic!("{}: Expectation({}) called {} time(s) which is fewer than expected {}",
+                                    #funcname,
+                                    desc,
+                                    self.times.count(),
+                                    self.times.minimum());
+                            },
+                            ::mockall::ExpectedCalls::SatisfiedOnlyLowerBound => {
+                                panic!("Exceeded number of expected calls panic propagation");
```
asomers, line 1001: What's this about panic propagation?  A better message would be:
```suggestion
                                panic!("{}: Expectation({}) called {} time(s) which is more than the permitted {}",
                                    #funcname,
                                    desc,
                                    self.times.count(),
                                    self.times.maximum() - 1);
```

mockall/src/lib.rs
```diff
@@ -1554,6 +1554,14 @@ impl From<RangeToInclusive<usize>> for TimesRange {
     }
 }
 
+#[derive(PartialEq)]
+#[doc(hidden)]
+pub enum ExpectedCalls {
+    Satisfied,
+    SatisfiedOnlyLowerBound,
+    SatisfiedOnlyUpperBound,
```
asomers, line 1562: I find this nomenclature confusing, because it focuses on which bounds are satisfied, rather than which aren't, which is more interesting.  I suggest
```suggestion
    Satisfied,
    TooFew,
    TooMany,
```

asomers (changes requested):

vikz95:
Now that the tests failures are fixed I need to write the 3 test cases:
1. actual calls < than expected
2. actual calls = expected
3. actual calls > expected

> No need to get all fancy with threads. Just use [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html#).
But if I don't use threads then in case 3 the call() method will panic, and not the drop().

mockall_derive/src/mock_function.rs
vikz95: This message is already printed by the panic in the call() method.
So in a subthread scenario there will be 2 equal panic messages.
I wanted to differentiate the second one.

Or did you expected to remove completely the panic in the call() method? And have only one in drop()?

mockall/src/lib.rs
vikz95: Fine, will update

asomers:
> Now that the tests failures are fixed I need to write the 3 test cases:
> 
>     1. actual calls < than expected
> 
>     2. actual calls = expected
> 
>     3. actual calls > expected
> 
> 
> > No need to get all fancy with threads. Just use [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html#).
> 
> But if I don't use threads then in case 3 the call() method will panic, and not the drop().

That's what `catch_unwind` is for.  It will catch the panic and allow the test to continue.

mockall_derive/src/mock_function.rs
asomers: Will it be?  The second message will only be printed if for some reason the user is ignoring panics.  In that case, he'll need all the info in the second one.

vikz95: when calls happen in subthread and are more than expected fitst there is the panic from the check in call() method.
but it's ignored by main thread.
then there is the panic from the check in drop() method that makes the test fail.

but even if the messages are the same and are printed 2 times in this scenario it's not a problem, so I changed the message as suggested.

vikz95:
cargo doc is failing:
```
error: Unsupported in this context
  --> mockall/src/examples.rs:42:5
   |
42 |     /// An implementation of a trait on a mocked struct
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not document `mockall`
```
I don't know if it's due to my changes, and what is the issue.

vikz95:
In the end I choose to use threads in the tests to reproduce the issue that I encountered, and also because I don't know how to use catch_unwind() to ignore the panic in call() and fail on the panic in drop() when calls > expectation

asomers:
You should learn how to use catch_unwind.  It's more direct than using threads.  I'll ensure that the master branch works.

vikz95:
Ok, I used `catch_unwind` and tests are definetely simpler.

asomers:
Ok, I fixed master.  The problem was a change in the new syn-2.0.9 crate.  Could you please rebase your PR now?

CHANGELOG.md
```diff
@@ -31,6 +31,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Static methods' expectations will now be cleared during a panic.
   ([#443](https://github.com/asomers/mockall/pull/443))
 
+- Before dropping the mock object an additional check will verify if the
+  actual number of times a method was called is above the expectation.
+  This is needed because if the original check, that happens at every call,
+  panics in a subthread, then the caller thread will ignore the panic and 
+  the test will succeed while should have failed.
```
asomers, line 38: Or to put it much more simply:
```suggestion
- The `checkpoint` method now works correctly even after a panic due to too many
  method calls.
```

mockall/src/lib.rs
```diff
@@ -1602,10 +1610,23 @@ impl Times {
         (self.range.0.end - self.range.0.start) == 1
     }
 
-    /// Has this expectation already been called the minimum required number of
-    /// times?
-    pub fn is_satisfied(&self) -> bool {
-        self.count.load(Ordering::Relaxed) >= self.range.0.start
+    /// Has this expectation already been called a number of times above the minimum,
+    /// below the maximum, both cases, or neither?
```
asomers, line 1614: Let me suggest some more natural English:
```suggestion
    /// Has this expectation already been called the expected number of times?
    /// If not, was it too many or too few?
```

mockall/tests/check_expectations_on_object_drop.rs
```diff
@@ -0,0 +1,54 @@
+// vim: tw=80
```
asomers, line 1: You don't need a new test program for this.  The reason that Mockall's tests are split into so many separate test programs is so that most of them will still compile successfully even if one does not.  And when investigating a compile failure, it helps to narrow the search for the problem.
For this PR, you aren't changing mockall's syntax at all.  So I suggest that you delete this file and delete the `expectation_satisfied` and `too_few_calls` methods, which are redundant with tests elsewhere.  Then, move `too_many_calls` into `tests/mock_struct.rs`, in the `checkpoint` module.

asomers (changes requested):

asomers (approved):

asomers:
Thanks for your contribution @vikz95 .

vikz95:
Thank you for the support and guidance.
Do you already know when a new version with the changes will be released?

asomers:
I'm planning to make a patch release soon.  But not a feature release, because I'm still not comfortable with the implementation of the concretize feature.  I still haven't found a non-hacky way to do it.

PR #473, opened 2023-03-21
Changed example to use a C function available also on Windows

vikz95:
`getuid()` is a clib function that's only available on UNIX systems, this pull request uses the `time()` function that's also available on Windows.

Now when running `cargo test` on Windows the command doesn't fail with error message:
```
error[E0601]: `main` function not found in crate `ffi`
  --> mockall\examples\ffi.rs:35:2
   |
35 | }
   |  ^ consider adding a `main` function to `mockall\examples\ffi.rs`
```

### DIFF
--- a/mockall/examples/ffi.rs
+++ b/mockall/examples/ffi.rs
@@ -1,7 +1,6 @@
 // vim: tw=80
 //! An example of unit testing involving mocked FFI functions
 
-use std::ptr;
 #[cfg(test)]
 use mockall::automock;
 use mockall_double::double;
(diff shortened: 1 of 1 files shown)


mockall/examples/ffi.rs
```diff
@@ -1,35 +1,39 @@
...
-    println!("My uid is {}", getuid() );
+    println!("Seconds since epoch: {}", time() );
 }
 
 
 #[test]
-fn getuid_test() {
-    let ctx = ffi::getuid_context();
+fn time_test() {
+    let seconds_from_epoch = std::time::SystemTime::now()
```
asomers, line 30: The example would be both clearer, and more distinct from regular behavior, if you simply return a fixed constant.

mockall/examples/ffi.rs
```diff
@@ -1,35 +1,39 @@
...
+use std::ptr;
 #[cfg(test)]
 use mockall::automock;
 use mockall_double::double;
 
 #[cfg_attr(test, automock)]
 pub mod mockable_ffi {
     extern "C" {
-        pub fn getuid() -> u32;
+        pub fn time(timer: *const u64) -> u64;
```
asomers, line 12:
```suggestion
        pub fn time(timer: *mut u64) -> u64;
```

asomers (changes requested):

asomers (approved):
LGTM.

PR #517, opened 2023-09-21
Remove default features from predicates dependency

grimerssy:
Predicates has default optional features this crate doesn't use. These features even include regex, which is a heavy dependency. This PR should slightly improve compile times for crates using mockall


asomers (changes requested):
Hey, this is great!  Thanks for noticing.  One minor problem, though.  Since we reexport predicates, this is technically a breaking change.  Therefore, it warrants mention in the CHANGELOG.  Could you please add it?  Users should be able to fix any breakages by listing predicates, with any required features, in their own Cargo.toml files.

asomers (approved):

PR #527, opened 2023-12-09
Less Mutex poisoning

### LINKED ISSUE #515: Reduce mutex poisoning
Mocked static methods contain an internal Mutex .  If a test case panics while accessing it (for example, by violating a `.times()` constraint, or by explicitly panicing in the `.returning` function), then the Mutex will be poisoned, and all future test cases trying to use the same static method will fail with `PoisonError`.

https://github.com/asomers/mockall/blob/cab609110bf1d31e12ca0efadd8c81a0f9744d20/mockall_derive/src/mock_function.rs#L1400
https://github.com/asomers/mockall/blob/cab609110bf1d31e12ca0efadd8c81a0f9744d20/mockall_derive/src/mock_function.rs#L1469

But the `Expectations` object guarded by this Mutex gets cleared on Drop anyway.  So we may as well clear any poison at the same time.  That requires a nightly compiler.
https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.clear_poison

asomers:
If a test case panics while holding a static method's internal Mutex locked, for example by violating a times() constraint, then clear the Mutex's poison during Drop.  We're going to delete all of its Expectations anyway, so there's no data to be poisoned.

Requires nightly.

Fixes #515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Raised MSRV to 1.70.0 to remove `lazy_static` dependency
   ([#550](https://github.com/asomers/mockall/pull/550))
 
+- No longer poison a Context object's internal `Mutex` when panicing.  This
+  requires the "nightly" feature.
+  ([#527](https://github.com/asomers/mockall/pull/527))
+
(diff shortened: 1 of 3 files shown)


asomers:
This PR depends on the https://github.com/rust-lang/rust/issues/96469 feature, which should be merged soon.

PR #531, opened 2023-12-13
Fix using #[concretize] on functions with bounded generic types

### LINKED ISSUE #530: Concretize rejects auto-trait bounds
The concretize attribute rejects methods that contain generics of the form "Trait + AutoTraits", which are object-trait safe

```rust
use mockall::{automock, concretize};

trait A {}

#[automock]
trait B {
    #[concretize]
    fn method<T: A + Send>(&self);
}

struct C;

impl A for C {}

impl B for C {
    fn method<T: A + Send>(&self) {
        println!("Call for method")
    }
}

fn main() {
    let _a: Box<dyn A + Send> = Box::new(C);
}
```

However, removing the "automock" attribute the code compiles, since `dyn A + Send` is a valid trait object.

For now, a workaround is to define another trait that has the required trait and auto-traits as super-traits with a blanket implementation:

The following code compiles:

```rust
use mockall::{automock, concretize};

trait A {}

trait ASend: A + Send {}

impl<T: A + Send> ASend for T {}

#[automock]
trait B {
    #[concretize]
    fn method<T: ASend>(&self);
}

struct C;

impl A for C {}

impl B for C {
    fn method<T: ASend>(&self) {
        println!("Call for method")
    }
}

fn main() {
    let _a: Box<dyn A + Send> = Box::new(C);
}

```

asomers:
It's usually possible to turn such types into trait objects.  We just need to emit extra parenthesis in the generated code.

Fixes #530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 
+## [ Unreleased ] - ReleaseDate
+
+### Fixed
+
+- Fixed using `#[mockall::concretize]` on functions whose generic types contain
+  trait bounds, yet are still object safe.
+  ([#531](https://github.com/asomers/mockall/pull/531))
+
 ## [ 0.12.0 ] - 2023-12-10
 
 ### Added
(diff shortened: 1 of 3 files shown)


asomers:
cc @glueball can you please test this change?

glueball:
Tested with 

```toml
[patch.crates-io]
mockall = { git = "https://github.com/asomers/mockall", branch = "concretize-bounds" }
```

It works! Awesome ;)

PR #534, opened 2023-12-15
Fix mocking functions that use raw identifiers for their names.

### LINKED ISSUE #533: Raw method identifiers stopped working with mockall `0.12.0`
Since `mockall` `0.12.0` the following

```rust
#[mockall::automock]
trait Foo {
    fn r#match(&self);
}
```

doesn't compile anymore:
```
error: custom attribute panicked
 --> src/main.rs:1:1
1 | #[mockall::automock]
  | ^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: `"expect_r#match"` is not a valid identifier
```

With `0.11.4` this worked just fine.

asomers:
This was a regression in 0.12.0, introduced by 60acf8e .

Fixes #533

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   trait bounds, yet are still object safe.
   ([#531](https://github.com/asomers/mockall/pull/531))
 
+- Fixed mocking methods that use raw identifiers for their names.  This was a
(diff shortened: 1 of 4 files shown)


asomers:
@tim3z could you test please?

tim3z:
Works like a charm, both in the minimal example and the full project I encountered that bug. Thanks for the super quick response!

PR #550, opened 2024-02-08
Use `const Mutex::new()` whenever possible

Expyron:


mockall_derive/src/mock_function.rs
```diff
@@ -1397,12 +1397,10 @@ impl<'a> ToTokens for ConcreteExpectationGuard<'a> {
         let (ei_ig, _, _) = e_generics.split_for_impl();
         let v = &self.f.privmod_vis;
         quote!(
-            ::mockall::lazy_static! {
-                #[doc(hidden)]
-                #v static ref EXPECTATIONS:
-                    ::std::sync::Mutex<Expectations #tg> =
-                    ::std::sync::Mutex::new(Expectations::new());
-            }
+            #[doc(hidden)]
```
asomers, line 1400: Can you also eliminate the lazy_static at line 1469? If so, we can eliminate lazy_static as a dependency.

asomers (commented):
Nice job.  Do you know what MSRV is required for this to work?

Expyron:
MSRV for `const Mutex::new()` is 1.63, which should already be OK for this crate.

mockall/Cargo.toml
```diff
@@ -53,7 +52,7 @@ mockall_double = { version = "^0.3.1", path = "../mockall_double" }
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
-tracing = "0.1.32"
+tracing = "0.1.35"
```
asomers, line 55: Why was it necessary to change this?

Expyron: A bug in `tracing` (actually its dependency `tracing-core`) below 0.1.35 when lazy_static version is too low.
It was raising an error in the CI when testing with `cargo update -Zminimal-versions`.

asomers (approved):

asomers:
Thanks a lot, @Expyron !

PR #555, opened 2024-02-24
Handle the #[inline], #[cold], and #[must_use] attributes

### LINKED ISSUE #552: Can't apply `automock` to traits with default method implementions marked `#[inline]`
The following code fails to compile:

```rust
#[mockall::automock]
trait Foo {
    #[inline]
    fn foo() -> i32 {
        1
    }
}

fn main() {}
```

With the error:

```
error[E0518]: attribute should be applied to function or closure
 --> src/main.rs:3:5
  |
1 | #[mockall::automock]
  | -------------------- not a function or closure
2 | trait Foo {
3 |     #[inline]
  |     ^^^^^^^^^
```

The issue is that the default method implementation is marked `#[inline]`. The same problem occurs for `#[inline(always|never)]`.

mockall: 0.12.1
rustc: 1.76.0

---

I also checked `#[cold]` and `#[must_use]`. `#[cold]` causes a warning to be emitted that will become an error in the future:

```
warning: attribute should be applied to a function definition
 --> src/main.rs:3:5
  |
1 | #[mockall::automock]
  | -------------------- not a function definition
2 | trait Foo {
3 |     #[cold]
  |     ^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: `#[warn(unused_attributes)]` on by default
```

`#[must_use]` causes a warning, but as of Rust 1.76.0, it's not scheduled to become a hard error (edit: oops, pasted in the wrong warning originally):

```
warning: `#[must_use]` has no effect when applied to a module
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^
  |
  = note: `#[warn(unused_attributes)]` on by default
```

There may be other attributes that cause problems, but these were all I tested.

asomers:
Fixes #552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [ Unreleased ] - ReleaseDate
 
+### Added
+
+- Add the ability to mock methods that use `#[inline]` or `#[cold]`.
+  ([#555](https://github.com/asomers/mockall/pull/555))
+
 ### Changed
 
 - Raised MSRV to 1.70.0 to remove `lazy_static` dependency
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Add the ability to mock methods that use `#[inline]` or `#[cold]`.
+- Add the ability to mock methods that use `#[inline]` or `#[cold]`, and
+  methods or traits that use `#[must_use]`.
   ([#555](https://github.com/asomers/mockall/pull/555))
 
 ### Changed
--- a/mockall/tests/automock_inline.rs
+++ b/mockall/tests/automock_inline.rs
@@ -1,0 +1,52 @@
+// vim: tw=80
+//! Mockall should ignore and not emit attributes like "inline" that affect
+//! code generation.
+#![deny(warnings)]
+
+use mockall::*;
+
+pub struct Foo {}
+
+#[automock]
+impl Foo {
+    #[inline]
+    pub fn foo(&self) -> i32 {unimplemented!()}
+    #[inline]
+    pub fn bar() -> i32 {unimplemented!()}
+    #[cold]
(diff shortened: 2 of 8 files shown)


asomers:
@tgnottingham does this branch work for you?

tgnottingham:
@asomers -- yes, it works for my use case! Thanks for the quick fix!

This hasn't affected me yet in practice, but I noticed that there are still some warnings emitted for `#[must_use]` (regardless of whether or not there's a default method implementation):

```rust
#[mockall::automock]
trait Foo {
    #[must_use]
    fn foo(&self) -> i32;
}

fn main() {}
```
On Rust 1.76.0:

```
warning: `#[must_use]` has no effect when applied to a struct field
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^
  |
  = note: `#[warn(unused_attributes)]` on by default

warning: `#[must_use]` has no effect when applied to a struct field
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^

warning: `#[must_use]` has no effect when applied to an expression
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^

warning: unused attribute
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^ help: remove this attribute
  |
note: attribute also specified here
 --> src/main.rs:1:1
  |
1 | #[mockall::automock]
  | ^^^^^^^^^^^^^^^^^^^^
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: this warning originates in the attribute macro `mockall::automock` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `#[must_use]` has no effect when applied to a module
 --> src/main.rs:3:5
  |
3 |     #[must_use]
  |     ^^^^^^^^^^^
```

asomers:
@tgnottingham I cannot reproduce those warnings.  Are you sure that you're using the "inline" branch?

tgnottingham:
My mistake, I was on the first commit to that branch. Everything works great on the second commit. Thanks again!

PR #572, opened 2024-04-27
Fix sequences doc test

CardboardTurkey:
`should_panic` doesn't accept an `expected` parameter for doc tests
(unlike unit tests).

Without this commit cargo doesn't recognise/render it as a doc test.


CardboardTurkey:
See [docs.rs](https://docs.rs/mockall/latest/mockall/#sequences) for how the doc test isn't currently rendered properly

asomers (approved):
Thanks for the tip!

PR #579, opened 2024-05-18
fix(mockall_derive): add allow(clippy::indexing_slicing) to generated module.

JosiahBull:
I tend to configure `Cargo.toml` with:

```toml
[workspace.lints.clippy]
indexing_slicing = "deny"
```

[https://rust-lang.github.io/rust-clippy/master/index.html#/indexing_slicing](https://rust-lang.github.io/rust-clippy/master/index.html#/indexing_slicing)

Among other lints to catch code that might panic. The generated code from `mockall` currently fails this lint, and it's difficult to add an exception for it with `#[allow()]` as it's occurring within a proc macro.

This PR updates codegen to replace the instance of index slicing with `.get().unwrap()`. This does not represent any real change in functionality, just passes the lint check now. It will still panic the same as it would have previously.

I am happy to change this to add a `#[allow()]` annotation in the generated code if this is preferred. :)

### DIFF
--- a/mockall_derive/src/mock_function.rs
+++ b/mockall_derive/src/mock_function.rs
@@ -860,7 +860,7 @@ impl MockFunction {
         quote!(
             #(#attrs)*
             #[allow(missing_docs)]
-            #[allow(clippy::too_many_arguments)]
+            #[allow(clippy::too_many_arguments, clippy::indexing_slicing)]
             pub mod #inner_mod_ident {
                 use super::*;
                 use ::mockall::CaseTreeExt;


mockall_derive/src/mock_function.rs
```diff
@@ -1211,7 +1211,7 @@ impl<'a> ToTokens for CommonExpectationsMethods<'a> {
                 {
                     self.0.push(Expectation::default());
                     let __mockall_l = self.0.len();
-                    &mut self.0[__mockall_l - 1]
+                    self.0.get_mut(__mockall_l - 1).unwrap()
```
asomers, line 1214: Isn't a `get_mut` followed by an `unwrap` equivalent to slicing?  It seems like it would just change one panic message for another.  A change that would at least improve readability would be to use `last_mut` instead.  That would be usable here and in the other locations too.
```suggestion
                    self.0.last_mut().unwrap()
```

asomers (changes requested):

mockall_derive/src/mock_function.rs
JosiahBull: >  It seems like it would just change one panic message for another.

You are correct, I use this lint to avoid implicit panics in code. With the `mockall` library a panic is desirable as it should only be run during tests. If you'd prefer not to have the additional `.unwrap()`'s added I'm happy to change this to add `#[allow(clippy::indexing_slicing)]` tags where appropriate, or change to `.expect()` calls.

The goal is to allow the proc-macro to compile for end-users with `#[deny(clippy::indexing_slicing)]` set on their projects, as right now it generates an error from inside the proc macro.

asomers: I don't think that replacing slicing with .get_mut offers any advantages to Mockall, so I don't want to do it.  But using last_mut() instead of slicing removes a local variable, and that's a good thing.  BTW, you can also use last_mut() in all of the other places that this PR affects; not just the one I pointed out.

JosiahBull: I understand the unwillingness to update the library - this doesn't change any functionality.

Do you have a recommendation on how better to address this issue? Not sure on how best to handle errors from inside a proc macro like this.

image

JosiahBull: You are right, this does not offer material changes to mockall. I'm unsure how other to resolve the error from inside the proc macro though: [https://pastebin.com/K8XTbuJ6](https://pastebin.com/K8XTbuJ6)

Do you have advice on how I can best resolve this in my crate?

Because there are multiple methods annotating with `#[allow()]` does not work. :(

image

asomers: It would be fine to add an `#[allow()]` to the generated code.  There are already several such allowances.

asomers (changes requested):
Instead of adding the `#[allow()]` attribute to every single function, could you add it just once for the private module?

asomers (approved):

asomers:
LGTM, thanks.

PR #585, opened 2024-06-22
Use the C-unwind ABI instead of C when mocking extern C functions

### LINKED ISSUE #584: extern C functions need to be C-unwind in Rust 1.81.0
A recent change to rustc (https://github.com/rust-lang/rust/issues/74990) changed the behavior of unwinding from a Rust function across an `extern C` ABI change.  Now, that will cause an abort.  This is causing Mockall's CI tests to fail.  The possible solutions are either:
* Require users to specify `extern C-unwind` instead.  This is annoying for users, and it might not be correct for all non-mock functions.
* Automatically define the mock functions as `C-unwind`.  This is much better.  However, it causes two new problems:
  - A function pointer to the original C function will now have a different type from a function pointer to the mock function, because the ABI is part of the function pointer's type.
  - `C-unwind` does not yet work for variadic functions.
* Somehow figure out how to make Mockall work with `panic=abort`.  It's not a problem, as long as no tests are `should_panic` and no tests ever fail :wink: .

asomers:
Begining within Rust 1.81
(https://github.com/rust-lang/rust/issues/74990) Rust will abort when attempting to unwind a panic across an "extern C" boundary.  Previously it was technically UB, but it worked and Mockall relied on it.  Now, unwinding across such a boundary requires the "extern C-unwind" ABI. Use that ABI in the generated code.

However, don't use that ABI when mocking a C variadic function.  That doesn't work, due to https://github.com/rust-lang/rust/issues/126836 .

Fixes #584

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,7 +14,7 @@ common: &COMMON
 task:
   name: MSRV
   container:
-    image: rust:1.70.0
+    image: rust:1.71.0
   cargo_lock_script:
     - cp Cargo.lock.msrv Cargo.lock
   << : *COMMON
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,18 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Raised MSRV to 1.70.0 to remove `lazy_static` dependency
-  ([#550](https://github.com/asomers/mockall/pull/550))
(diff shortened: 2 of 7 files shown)


asomers:
@Enes1313 since you are the one who first requested the `extern "C"` linkage, could you please test your project with this PR branch?

asomers:
@Enes1313 I'm going to go ahead and merge this since CI on the master branch is broken .  But I'd still appreciate any post-merge feedback you can provide.

PR #587, opened 2024-06-24
Use the C-unwind ABI when mocking variadic functions

asomers:
Now that https://github.com/rust-lang/rust/issues/126836 is fixed.


asomers:
This PR is failing because the rustlang/rust:nightly container image is too old.  Hopefully it will be updated soon.

PR #589, opened 2024-07-05
Fix autolinks in the documentation

Jujumba:


mockall/src/lib.rs
```diff
@@ -1368,7 +1368,7 @@ pub use mockall_derive::concretize;
 /// }
 /// # fn main() {}
 /// ```
-/// Mocking an unsupported `#[derive(X)]` attribute, e.g. `Clone`, is
+/// Mocking an unsupported `#[derive(X)]` attribute, e.g. [`Clone`], is
```
asomers, line 1371: Does this link actually work?  Did you generate the docs to check?  And is it even worthwhile to make a link here? I don't think many readers will need one.

mockall/src/lib.rs
```diff
@@ -1388,7 +1388,7 @@ pub use mockall_derive::concretize;
 ///
 /// When mocking a generic struct's implementation of a generic trait, use the
 /// same name for their generic parameters.  For example, if you wanted to mock
-/// `Rc`, do
+/// [`Rc`], do
```
asomers, line 1391: Likewise, I don't think this needs to be a link.  It's not really important for this example that users know exactly what Rc does.

mockall/src/lib.rs
```diff
@@ -1273,7 +1273,7 @@ pub mod examples;
 /// # Limitations
 ///
 /// `#[automock]` can't handle everything.  There are some cases where
-/// you will need to use [`mock`] instead:
+/// you will need to use [`mock!`] instead:
```
asomers, line 1276: The link works as-is.  Does it still work after this change?  Did you generate the docs to check?

mockall/src/lib.rs
Jujumba: Yes, the links are valid. I've added them for the sake of consistency

mockall/src/lib.rs
Jujumba: > The link works as-is. Does it still work after this change? Did you generate the docs to check?

Indeed I've checked them ;)

asomers (changes requested):
I think you should remove the link to std::rc::Rc;  It isn't helpful here.

asomers (approved):

Issue #600, opened 2024-08-12
Creating a mock with multiple traits fails if traits share a name


discosultan:
In mockall = "0.13":

```rs
use mockall::mock;

mod a {
    pub trait A {
        fn foo(&self);
    }
}

mod b {
    pub trait A {
        fn bar(&self);
    }
}

mock! {
    // Structure to mock
    C {}
    // First trait to implement on C
    impl a::A for C {
        fn foo(&self);
    }
    // Second trait to implement on C
    impl b::A for C {
        fn bar(&self);
    }
}
```

fails with:

```
error[E0428]: the name `__mock_MockC_A` is defined multiple times
  --> src/lib.rs:15:1
   |
15 | mock! {
   | ^^^^^
   | |
   | `__mock_MockC_A` redefined here
   | previous definition of the module `__mock_MockC_A` here
   |
   = note: `__mock_MockC_A` must be defined only once in the type namespace of this module
   = note: this error originates in the macro `mock` (in Nightly builds, run with -Z macro-backtrace for more info)
```

asomers:
Ahh yes.  That's bound to cause problems. I bet if you expand the code, you'll find more than just those two identically named structs, too.  This limitation is unlikely to be fixed.  But if it's really important to you, you could take a stab at it.  Perhaps you could name the structs using the full path to the trait, replacing `::` with `__` ?

discosultan:
I gave it a go in https://github.com/asomers/mockall/pull/601.

asomers:
Fixed by #601

PR #601, opened 2024-08-12
Support mocking multiple traits with same name

discosultan:
An attempt to solve https://github.com/asomers/mockall/issues/600.

Not really sure about backward compatibility. Should be fine as we only change the generation of a private identifier. But it could cause issues if either a length limit is reached or a there's a name collision with a different ident.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [ Unreleased ] - ReleaseDate
+
+### Fixed
+
+- Fixed naming conflict when mocking multiple traits with same name but from
+  different modules.
+  ([#601](https://github.com/asomers/mockall/pull/601))
+
 ## [ 0.13.0 ] - 2024-07-21
 
 ### Added
--- a/mockall_derive/src/mock_trait.rs
+++ b/mockall_derive/src/mock_trait.rs
@@ -3,6 +3,7 @@ use proc_macro2::Span;
 use quote::{ToTokens, format_ident, quote};
 use std::{
    collections::hash_map::DefaultHasher,
+   fmt::Write,
    hash::{Hash, Hasher}
 };
 use syn::{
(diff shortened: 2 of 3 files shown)


mockall_derive/src/mock_trait.rs
```diff
@@ -176,3 +174,45 @@ impl MockTrait {
         )
     }
 }
+
+fn write_join_path(
```
asomers, line 178: It would be more idiomatic if this function allocated and returned a new String.

mockall_derive/src/mock_trait.rs
```diff
@@ -176,3 +174,45 @@ impl MockTrait {
...
+        path_args.hash(&mut hasher);
+        is_some = true;
+    }
+    is_some.then(|| hasher.finish())
+}
+
+// Similar to how the `format_ident!` constructs an `Ident`.
+// https://github.com/dtolnay/quote/blob/3256f897c9821f9b397cb81851ae9a1672a84cb7/src/runtime.rs#L478-L484
+fn ident_maybe_raw(id: &str, span: Span) -> Ident {
+    if let Some(id) = id.strip_prefix("r#") {
```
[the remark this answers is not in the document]
discosultan: It caused the existing [raw_identifier](https://github.com/asomers/mockall/blob/63a6b76054a160c51554839097315b98c788f3fc/mockall/tests/raw_identifier.rs) test to not compile

mockall_derive/src/mock_trait.rs
```diff
@@ -176,3 +174,45 @@ impl MockTrait {
...
+            is_first = false;
+            write!(output, "{}", segment.ident)?;
+        } else {
+            write!(output, "{}{}", sep, segment.ident)?;
+        }
+    }
+    Ok(())
+}
+
+fn hash_path_arguments(path: &Path) -> Option<u64> {
```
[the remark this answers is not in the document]
discosultan: I don't have an example. This was done since I assumed processing the hash + concatenating the ident are symmetric. Whereas before both were done on the last segment, now we go over all the segments. Happy to revert this part to only process the last segment.

mockall_derive/src/mock_trait.rs
discosultan: Will update. I was trying to avoid string allocations but then again, I couldn't figure out a way to know the required capacity ahead of time anyway, so even the current single `String::new` approach keeps resizing it internally.

mockall_derive/src/mock_trait.rs
```diff
@@ -176,3 +174,35 @@ impl MockTrait {
         )
     }
 }
+
+fn join_path_segments(path: &Path, sep: &str) -> String {
+    let mut output = String::new();
+    for segment in &path.segments {
+        if write!(
```
asomers, line 181: Too bad there isn't something like a `Punctuated::join` method to help with this part.

asomers (changes requested):
Looks good.  Now how about a CHANGELOG entry?

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [ unreleased ]
```
asomers, line 6:
```suggestion
## [ Unreleased ] - ReleaseDate
```

asomers (changes requested):

asomers (approved):
Thanks for your contribution!

PR #640, opened 2025-02-11
Add support for trait-variant

### LINKED ISSUE #639: Compatibility with trait-variant for automock
This crate _automock_ already has direct support for _async_trait_ (#183), allowing you to place the _async_trait_ macro after the _automock_ ([doc](https://docs.rs/mockall/latest/mockall/#async-traits)) which allows you to directly return the future output when mocking:

```rs
#[automock]
#[async_trait]
pub trait Foo {
    async fn foo(&self) -> i32;
}

#[tokio::test]
async fn test() {
    mock.expect_foo().return_const(1);
    assert_eq!(1, mock.foo().await);
}
```

If you place the _async_trait_ to be resolved before, you have to return the pinned box future on the mocks, which allows for further customization:

```rs
#[async_trait]
#[automock]
pub trait Foo {
    async fn foo(&self) -> i32;
}

#[tokio::test]
async fn test() {
    mock.expect_foo().returning(|| Box::pin(async { 1 }));
    assert_eq!(1, mock.foo().await);
}
```

That's great! And I was hoping now that [_trait-variant_](https://crates.io/crates/trait-variant) is available as part of Rust 1.75 [stabilization of _async fn_ in traits](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html) that it could work the same way.

This is working:

```rs
#[trait_variant::make(Send)]
#[automock]
pub trait Foo {
    async fn foo(&self) -> i32;
}

#[tokio::test]
async fn test() {
    mock.expect_foo().returning(|| Box::pin(async { 1 }));
    assert_eq!(1, mock.foo().await);
}
```

But this is failing:

```rs
#[automock]
#[trait_variant::make(Send)]
pub trait Foo {
    async fn foo(&self) -> i32;
}

#[tokio::test]
async fn test() {
    mock.expect_foo().return_const(1);
    assert_eq!(1, mock.foo().await);
}
```

If the trait variant annotation is removed it works as expected, and if the annotation is added back to the expanded trait it works fine also, so I guess this is not a big change, but I'm not familiar with the codebase.

It would be great to support it the same way in order to allow for an easy migration from _async_trait_ to _trait_variant_, without having to _fix_ all of the tests.

lasantosr:
This PR aims to fix #639 

Feel free to update the readme, as English is not my primary language

### DIFF
--- a/mockall/src/lib.rs
+++ b/mockall/src/lib.rs
@@ -1067,15 +1067,14 @@
 //!
 //! ## Async Traits
 //!
-//! Async traits aren't yet (as of 1.47.0) a part of the Rust language.  But
-//! they're available from the
-//! [`async_trait`](https://docs.rs/async-trait/0.1.38/async_trait/) crate.
-//! Mockall is compatible with this crate, with two important limitations:
-//!
-//! * The `#[automock]` attribute must appear _before_ the `#[async_trait]`
-//! attribute.
-//!
-//! * The `#[async_trait]` macro must be imported with its canonical name.
+//! Partial support for async traits was introduced in the Rust language since
+//! 1.75.0.
+//! Mockall is compatible with them, as well as both 
+//! [`async_trait`](https://docs.rs/async-trait/latest/async_trait/) and
+//! [`trait_variant`](https://docs.rs/trait-variant/latest/trait_variant/)
+//! crates, with an important limitation:
+//! 
+//! * The `#[automock]` attribute must appear _before_ the crate's attribute.
 //!
 //! ```
 //! # use async_trait::async_trait;
--- a/mockall/tests/automock_trait_variant.rs
+++ b/mockall/tests/automock_trait_variant.rs
@@ -1,0 +1,30 @@
+// vim: tw=80
+//! An async trait, for use with Futures
+#![deny(warnings)]
+
+use futures::executor::block_on;
+use mockall::*;
+
+#[automock]
+#[trait_variant::make(Send)]
+pub trait Foo {
+    async fn foo(&self) -> u32;
+    async fn bar() -> u32;
+}
+
+
+#[test]
+fn return_const() {
+    let mut mock = MockFoo::new();
+    mock.expect_foo()
+        .return_const(42u32);
+    assert_eq!(block_on(mock.foo()), 42);
+}
+
+#[test]
+fn static_method() {
+    let ctx = MockFoo::bar_context();
+    ctx.expect()
+        .return_const(42u32);
+    assert_eq!(block_on(MockFoo::bar()), 42);
+}
--- a/mockall/tests/mock_trait_variant.rs
+++ b/mockall/tests/mock_trait_variant.rs
@@ -1,0 +1,27 @@
+// vim: tw=80
+//! An async trait, for use with Futures
+#![deny(warnings)]
+
+use futures::executor::block_on;
+use mockall::*;
+
+#[trait_variant::make(Send)]
+pub trait Foo {
+    async fn foo(&self) -> u32;
+}
+
+mock! {
+    pub Bar { }
+    #[trait_variant::make(Send)]
+    impl Foo for Bar {
+        async fn foo(&self) -> u32;
+    }
+}
+
+#[test]
+fn return_const() {
+    let mut mock = MockBar::new();
+    mock.expect_foo()
+        .return_const(42u32);
+    assert_eq!(block_on(mock.foo()), 42);
+}
--- a/mockall_derive/src/lib.rs
+++ b/mockall_derive/src/lib.rs
@@ -771,6 +771,7 @@ fn find_lifetimes(ty: &Type) -> HashSet<Lifetime> {
 struct AttrFormatter<'a>{
     attrs: &'a [Attribute],
     async_trait: bool,
+    trait_variant: bool,
     doc: bool,
     must_use: bool,
 }
@@ -780,6 +781,7 @@ impl<'a> AttrFormatter<'a> {
         Self {
             attrs,
             async_trait: true,
+            trait_variant: false,
             doc: true,
             must_use: false,
         }
@@ -787,6 +789,11 @@ impl<'a> AttrFormatter<'a> {
 
     fn async_trait(&mut self, allowed: bool) -> &mut Self {
         self.async_trait = allowed;
+        self
+    }
+
+    fn trait_variant(&mut self, allowed: bool) -> &mut Self {
+        self.trait_variant = allowed;
         self
     }
 
@@ -820,6 +827,8 @@ impl<'a> AttrFormatter<'a> {
                     self.doc
                 } else if *i.as_ref().unwrap() == "async_trait" {
                     self.async_trait
+                } else if *i.as_ref().unwrap() == "make" {
+                    self.trait_variant
                 } else if *i.as_ref().unwrap() == "expect" {
                     // This probably means that there's a lint that needs to be
                     // surpressed for the real code, but not for the mock code.
--- a/mockall/Cargo.toml
+++ b/mockall/Cargo.toml
@@ -46,6 +46,7 @@ predicates-tree = "1.0"
 mockall_derive = { version = "=0.13.1", path = "../mockall_derive" }
 
 [dev-dependencies]
+trait-variant = "0.1.2"
 async-trait = "0.1.38"
 auto_enums = "0.8.5"
 futures = "0.3.7"
--- a/mockall_derive/src/mockable_struct.rs
+++ b/mockall_derive/src/mockable_struct.rs
@@ -371,6 +371,7 @@ impl From<(Attrs, ItemTrait)> for MockableStruct {
         let mut attrs = AttrFormatter::new(&trait_.attrs)
             .doc(true)
             .async_trait(true)
+            .trait_variant(true)
             .must_use(false)
             .format();
         attrs.push(derive_debug());


lasantosr:
@asomers I think the nightly is failing because some sort of new lints unrelated to this PR, would you like me to fix those also here?

asomers:
> @asomers I think the nightly is failing because some sort of new lints unrelated to this PR, would you like me to fix those also here?

It's already fixed on master.  You just need to rebase.

lasantosr:
Great! It looks like it's ready now

mockall/src/lib.rs
```diff
@@ -1067,15 +1067,14 @@
...
 //!
-//! Async traits aren't yet (as of 1.47.0) a part of the Rust language.  But
-//! they're available from the
-//! [`async_trait`](https://docs.rs/async-trait/0.1.38/async_trait/) crate.
-//! Mockall is compatible with this crate, with two important limitations:
-//!
-//! * The `#[automock]` attribute must appear _before_ the `#[async_trait]`
-//! attribute.
-//!
-//! * The `#[async_trait]` macro must be imported with its canonical name.
```
asomers, line 1078: This line is important, too.  You can't do something like `use async_trait::async_trait as something_else;`  Please restore this line.

mockall/tests/automock_trait_variant.rs
```diff
@@ -0,0 +1,30 @@
+// vim: tw=80
+//! An async trait, for use with Futures
```
asomers, line 2: This line is copy/pasted.  Please fix it.

mockall/tests/mock_trait_variant.rs
```diff
@@ -0,0 +1,27 @@
+// vim: tw=80
+//! An async trait, for use with Futures
```
asomers, line 2: This line is copy/pasted.  Please fix it.

asomers (changes requested):
I think the tests could really benefit from some kind of assertion to check that they actually implement the additional trait.  That might be hard to do with `Send`, but what about something like this?
```rust
trait Foo {}
#[automock]
#[trait_variant::make(Foo)]
pub trait Bar {
    ...
}
#[test]
fn some_test_method_name() {
    let mock = MockBar::new();
    let p: Box<dyn Foo> = Box::new(mock);   // Ensures that MockBar implements Foo
    ...
}
```
Also, with this change, does the `#[trait_variant::make(OtherTraitName: Foo)]` syntax work?

mockall/src/lib.rs
lasantosr: I totally missed that one, thanks!

lasantosr:
> I think the tests could really benefit from some kind of assertion to check that they actually implement the additional trait. That might be hard to do with `Send`, but what about something like this?
> 
> ```rust
> trait Foo {}
> #[automock]
> #[trait_variant::make(Foo)]
> pub trait Bar {
>     ...
> }
> #[test]
> fn some_test_method_name() {
>     let mock = MockBar::new();
>     let p: Box<dyn Foo> = Box::new(mock);   // Ensures that MockBar implements Foo
>     ...
> }
> ```
> 
> Also, with this change, does the `#[trait_variant::make(OtherTraitName: Foo)]` syntax work?

I'm not that familiar with the codebase to implement all of those changes.

On your first example, _automock_ would not implement `Foo` trait for `MockBar`, because it might have any arbitrary number of unrelated methods, but it could be manually implemented:


```rust
trait Foo {}
#[automock]
#[trait_variant::make(Foo)]
pub trait Bar {
    ...
}
impl Foo for MockBar {}
```

The most common use case is to mark your trait (and async method futures) `Send`, and mocks are already _Send_ so that's covered.

As for the second example, the syntax already works, but we would be mocking the initial trait only, not the other variant (with that syntax you're actually defining two traits). 

The main goal of this PR is to have feature parity with _async_trait_, other improvements could be added later.

As mentioend on #639 the _async_trait_ annotation can be located before or after the _automock_ annotation, and that means you need to return the actual pinned future or just the future's output on the mocks expectations.

With this PR we can do the same for _trait_variant_, so that the migration is painless.

lasantosr:
@asomers can this PR be merged as is to allow initial basic support for _trait_variant_?

Other improvements can be added later, when needed

mockall/src/lib.rs
```diff
@@ -1067,15 +1067,17 @@
...
+//! Partial support for async traits was introduced in the Rust language since
+//! 1.75.0.
+//! Mockall is compatible with them, as well as both 
+//! [`async_trait`](https://docs.rs/async-trait/latest/async_trait/) and
+//! [`trait_variant`](https://docs.rs/trait-variant/latest/trait_variant/)
+//! crates, with an important limitation:
```
asomers, line 1075:
```suggestion
//! crates, with two important limitations:
```

mockall/tests/automock_trait_variant.rs
```diff
@@ -0,0 +1,29 @@
...
+use futures::executor::block_on;
+use mockall::*;
+
+#[automock]
+#[trait_variant::make(Send)]
+pub trait Foo {
+    async fn foo(&self) -> u32;
+    async fn bar() -> u32;
+}
+
```
asomers, line 14: You should add a static assertion here to make sure that MockFoo really is `Send`.  You can use https://docs.rs/static_assertions/latest/static_assertions/macro.assert_impl_all.html .

mockall/tests/mock_trait_variant.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+mock! {
+    pub Bar { }
+    #[trait_variant::make(Send)]
+    impl Foo for Bar {
+        async fn foo(&self) -> u32;
+    }
+}
+
```
asomers, line 20: And add a similar assertion here.

asomers (changes requested):
We need to decide what do do about the `#[trait_variant::make(IntFactory: Send)]` syntax.  It might just work.  Have you tried it?  If so, we should add tests for it.  If not, and it won't be easy to do, then you need to at least document that.
Also, there must be a CHANGELOG entry.

mockall/tests/automock_trait_variant.rs
lasantosr: This is not required, as it wouldn't even compile if `MockFoo` is not `Send`.

The trait is expanded to:
```rs
pub trait Foo: Send { ... }
```

Meaning any type implementing `Foo` must also implement `Send`. Automock expands to:

```rs
impl Foo for MockFoo { ... }
```

It would not compile if `MockFoo` doesn't implement `Send`.

----

For example, this code does not compile:
```rs
trait Bar {}
#[automock]
#[trait_variant::make(Bar)]
pub trait Foo {
    ...
}
```

`Bar` trait would have to be manually implemented for `MockFoo` in order for it to compile.

lasantosr:
> We need to decide what do do about the `#[trait_variant::make(IntFactory: Send)]` syntax. It might just work. Have you tried it? 

@asomers It was already working but it mocks the original trait (not the variant)

On this example, `#[automock]` mocks `LocalFoo`, generating `MockLocalFoo`:
```rs
#[automock]
#[trait_variant::make(Foo: Send)]
pub trait LocalFoo {
    async fn foo(&self) -> u32;
    async fn bar() -> u32;
}
```

I've extended _automock_ to allow mocking the variant as well (I've included examples and updated _automock_ documentation)

On this example, `#[automock(target = Foo)]` mocks `Foo`, generating `MockFoo`:
```rs
#[automock(target = Foo)]
#[trait_variant::make(Foo: Send)]
pub trait LocalFoo {
    async fn foo(&self) -> u32;
    async fn bar() -> u32;
}
```

**Note:** `trait-variant` automatically generates a blanket impl for any type implementing `Foo` to also implement `LocalFoo`, so the resulting `MockFoo` can be used anywhere expecting both `LocalFoo` or `Foo`

All of the other comments have been resolved, except for the static assertion which is not required. If you still want me to include it let me know.

mockall/src/lib.rs
```diff
@@ -1231,6 +1233,17 @@ pub mod examples;
 /// let mock = MockBar::new();
 /// ```
 ///
+/// Mocking a trait variant with a different name requires adding a metaitem to the
+/// attribute in order to mock the variant instead of the original:
```
asomers, line 1237: Could you please clarify in the docs what automock will do in this instance?  Will it create a `MockFoo`, `MockLocalFoo`, or both?  I think the first one, but the docs should be explicit.

mockall/tests/automock_trait_variant.rs
asomers: Maybe so, but I want to make it explicit.  I want to see the static assertion so nobody reading this test will have to understand how `trait_variant` works, as well as to guard against future changes to the trait_variant crate.

mockall_derive/src/lib.rs
```diff
@@ -823,6 +830,8 @@ impl<'a> AttrFormatter<'a> {
                     self.doc
                 } else if *i.as_ref().unwrap() == "async_trait" {
                     self.async_trait
+                } else if *i.as_ref().unwrap() == "make" {
```
asomers, line 833: It's unfortunate that the most significant word in this attribute name isn't the last one "make", but the first "trait_variant".  That's unlike every other attribute that we process.  And we may have to alter our processing in the future, if another crate defines an attribute called "make".  But for now we can leave it.

asomers (changes requested):
This is looking pretty good now.  But I do want to use static_assertions, like I mentioned.

lasantosr:
@asomers It should be ready now! Thanks for the review

asomers (changes requested):
That looks much better.  I have just one more request: could you please add a CHANGELOG entry?

lasantosr:
@asomers There's already a changelog entry, let me know if you would like another sentence.

asomers:
> @asomers There's already a changelog entry, let me know if you would like another sentence.

Of course you did.  But yesterday I was only viewing "changes since your last review" so I didn't see it.  My bad.

asomers (approved):

Issue #568, opened 2024-04-16
Feature request: support for `const generics` in `mock!` and `automock`


korbinian-knott-bl:
Hi,
first of all, we this crate extensively, it makes testing and mocking easy and accessible. Thanks for that!

We now hit one limitation though: `const generics` are not yet supported as the macro tells us: `message: #automock does not yet support generic constants.`

For instance, if you have a struct with an array

```rust
struct Status<const N: usize> {
    elements: [u32; N],
}
```

neither `automock` nor `mock!` for a `Status<const N: usize>` works as expected.

Is the support of `const generics` planned or are there any major roadblocks preventing the implementation?

If it's just a matter of resources I'd like to try and implement this even (if you'd be willing to give some guidance of course).

asomers:
More fundamentally, mocking fields isn't even possible.  How would you propose accessing the mock object's `elements` array?

asomers:
I think it could be done.  But just as you must for associated types. you'd have to tell automock what value of N to use.

korbinian-maier-bl:
So something like `#[automock(const N=5;)]` or whatever keyword you'd choose for the const generic, right?

astapleton:
+1 on this. I love using automock, but I frequently run up against the limitation on not being able to use const generics in trait definitions

asomers:
BTW, while Mockall doesn't currently support const generics, it does support associated constants.  So you might be able to refactor your code to use those, instead.  See https://github.com/asomers/mockall/blob/master/mockall/tests/automock_associated_const.rs for an example.

HaoYang670:
Hi, any process on this, I am blocking on the const generic too

lindjwhite:
Any new progress on this? Also in need of support for const generics.

jonatanzeidler:
I tried to add support in #649, keeping the mocked implementation generic over the const, just as for other generics. The current implementation only works for usize. Feel free to help with the questions in the PR.

jonatanzeidler:
I changed implementation in #649 to work with all kinds of const generics. Testers welcome!

jonatan-zeidler-bl:
This is supported now in 0.14.0.

PR #649, opened 2025-04-14
Allow const generics in mocks

jonatanzeidler:
This adds support for #568. 

Example that works with this change:

```rust
#[mockall::automock]
trait MyTrait<T, const C: usize> {
    fn my_fun(&self, val: [T; C]) -> [T; C];
}

fn main() {
    let mut mock = MockMyTrait::new();

    mock.expect_my_fun().returning(|v| v);
    assert_eq!(mock.my_fun([1; 3]), [1; 3]);
}
```


mockall_derive/src/mock_item_struct.rs
```diff
@@ -51,10 +51,11 @@ fn phantom_fields(generics: &Generics) -> Vec<TokenStream> {
...
                 )
             },
-            syn::GenericParam::Const(_) => {
-                compile_error(param.span(),
-                    "#automock does not yet support generic constants");
-                None
+            syn::GenericParam::Const(cp) => {
+                let ty = &cp.ident;
+                Some(
+                quote!(#phident: ::std::marker::PhantomData<[(); #ty]>)
```
jonatanzeidler, line 57: This only works with `usize`. Any idea how to put any const generic in `PhantomData`?

jonatanzeidler: The best I can come up with is a constant like:

```rust
impl<T, const C: u8> MockMyTrait<T, C> {
    const _PHANTOM1: u8 = C;
}
```

jonatanzeidler, line 57: Maybe a stupid question, but why do we need the phantom data in the first place? Should the expectations member not be enough to have all generic being used?

jonatanzeidler: Nevermind. Generics on a trait may be unused, but they must be used in a struct...

jonatanzeidler: I solved it by using the const generics type instead of phantom data and assign it with the actual value. This introduces a minor memory overhead on the mock struct which is of the size of the const generic. But it works with all types of const generics.

asomers:
This looks promising!  Could you add some integration tests in the mockall/tests directory?  I think that this feature probably doesn't require doctests, since no special syntax is required by the user.

mockall/tests/automock_const_generic_trait.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn bar(&self) -> [u8; L];
+}
+
+#[test]
+fn generic_arguments() {
+    let mut mock = MockA::<16usize, true>::new();
+    mock.expect_foo()
+        .with(mockall::predicate::eq([42_u8; 16usize]))
+        .once()
+        .returning(|_| ());
```
asomers, line 19:
```suggestion
        .return_const(());
```

mockall/tests/automock_const_generic_trait.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn foo(&self, t: [u8; L]);
+    fn bar(&self) -> [u8; L];
+}
+
+#[test]
+fn generic_arguments() {
+    let mut mock = MockA::<16usize, true>::new();
+    mock.expect_foo()
+        .with(mockall::predicate::eq([42_u8; 16usize]))
+        .once()
```
asomers, line 18: You can delete this line.  It isn't really relevant for this test.
```suggestion
```

mockall/tests/automock_const_generic_trait.rs
```diff
@@ -0,0 +1,28 @@
...
+        .returning(|_| ());
+    mock.foo([42_u8; 16usize]);
+}
+
+#[test]
+fn generic_return() {
+    let mut mock = MockA::<1_usize, false>::new();
+    mock.expect_bar().returning(|| [13_u8]);
```
asomers, line 26: `returning` certainly works, but `return_const` is more efficient, when it's possible to use.
```suggestion
    mock.expect_bar().return_const([13_u8]);
```

mockall/tests/automock_const_generic_struct.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub fn foo(&self, _t: [u8; L]) {}
+    pub fn bar(&self) -> [u8; L] { [0; L] }
+}
+
+#[test]
+fn generic_arguments() {
+    let mut mock = MockGenericStruct::<16usize, true>::new();
+    mock.expect_foo()
+        .with(mockall::predicate::eq([42_u8; 16usize]))
+        .once()
```
asomers, line 22:
```suggestion
```

mockall/tests/automock_const_generic_struct.rs
```diff
@@ -0,0 +1,32 @@
...
+        .returning(|_| ());
```
asomers, line 23:
```suggestion
        .return_const(());
```

mockall/tests/automock_const_generic_struct.rs
```diff
@@ -0,0 +1,32 @@
...
+    mock.foo([42_u8; 16usize]);
+}
+
+#[test]
+fn generic_return() {
+    let mut mock = MockGenericStruct::<1_usize, false>::new();
+    mock.expect_bar().returning(|| [13_u8]);
```
asomers, line 30:
```suggestion
    mock.expect_bar().return_const([13_u8]);
```

asomers (changes requested):
This looks pretty cool.  Could you add a test for a static method?  That is, one without a `&self` argument?  And if you do that, then it would also be good to have a negative test.  That is, a test which calls the static method with the correct argument values, but the wrong generic constant.  Should a test should be marked `#[should_panic]`.

jonatanzeidler:
Will add those tests, still.

jonatanzeidler:
Calling the function with the wrong const generic value basically means that there is no implementation for it. This is a compile time error, not a panic:
```rust
33 |     assert_eq!([13_u8], MockB::<2_usize, false>::bar());
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `[u8; 1] == [u8; 2]`
```

How would I add a test for that?

mockall/tests/automock_const_generic_struct_with_static_method.rs
```diff
@@ -0,0 +1,32 @@
...
+//! static non-generic methods of generic structs shouldn't require any special
+//! treatment when mocking.
+#![deny(warnings)]
+
+use mockall::*;
+
+// TODO: The version with just one const generic does not compile. ¯\_(ツ)_/¯
+
+//#[automock]
+//trait A<const L: usize> {
```
jonatanzeidler, line 11: For some reason, this does not compile when there is exactly one const generic argument, but it does compile when there are two or more. Do mind taking a look why this happens?

mockall/tests/automock_const_generic_struct_with_static_method.rs
```diff
@@ -0,0 +1,45 @@
...
+    assert_eq!([13_u8], MockB::<1_usize, false>::bar());
+}
+
+// TODO having two tests setting expectations on MockB::bar_context() makes
+// generic_return_2() crash. Is that expected?
+
+#[should_panic]
+#[test]
+fn wrong_const_generic_expectation() {
+    let ctx = MockB::<2_usize, false>::bar_context();
```
jonatanzeidler, line 41: Is it expected to be an issue to have two tests creating a `bar_context()`? This makes `generic_return_2` fail with a `PoisonError`.

asomers: It's fine for two tests to both call `bar_context`.  But to deconflict them, you should create a global Mutex.  See for example automock_generic_static_method.rs .

mockall/tests/automock_const_generic_struct_with_static_method.rs
asomers: I see the invalid generated code.  Something is generating a `L` when it should be a `()`.  Pro tip: you can see your generated code like this, and find out exactly which line is failing to compile:
```
$ cd mockall
$ sh tools/gen.sh automock_const_generic_struct_with_static_method
$ cp pp/automock_const_generic_struct_with_static_method.rs tests/t.rs
$ # Maybe copy some trait or struct definitions into tests/t.rs
$ cargo test --all-features --test t
```

jonatanzeidler: Thanks. That is useful. I was using rust-analyzer's expandMacro before, which wouldn't give me in-line error messages.

jonatanzeidler: I got the issue fixed in the key generation.

mockall/tests/automock_const_generic_struct_with_static_method.rs
jonatanzeidler: Making them run in serial doesn't fix the issue, because the panicking test leaves the static context Mutex poisoned. I moved the negative test to its own integration test.

mockall_derive/src/lib.rs
```diff
@@ -1028,7 +1028,7 @@ fn supersuperfy_bounds(
 fn gen_keyid(g: &Generics) -> impl ToTokens {
     match g.params.len() {
         0 => quote!(<()>),
-        1 => {
+        1 if matches!(g.params.first(), Some(GenericParam::Type(_))) => {
```
jonatanzeidler, line 1031: Please check whether this is a clean fix for the problem at hand.

mockall/tests/automock_const_generic_struct_with_static_method.rs
jonatanzeidler: Seems it was fixed in https://github.com/asomers/mockall/commit/204456a98e00b6d14d5daeb2752d68e1cf8a55b9 which I only got into this PR lately during a rebase. I moved the panicking test back in.

mockall/tests/automock_const_generic_struct_with_static_method.rs
```diff
@@ -0,0 +1,35 @@
...
+static A_MTX: Mutex<()> = Mutex::new(());
+
+#[automock]
+trait A<const L: usize, const C: bool> {
+    fn bar() -> [u8; L];
+}
+
+#[test]
+fn generic_return() {
+    let _m = A_MTX.lock().unwrap_or_else(PoisonError::into_inner);
```
asomers: Why the `PoisonError::into_inner`?  Also, this file mocks a `trait` not a `struct`, so the file name is wrong.

asomers (changes requested):

asomers (approved):

PR #668, opened 2025-10-09
Improve error message when call count is not specified in sequence

katshun0307:
The error message "Only Expectations with an exact call count have sequences" is misleading.
Changed it to "Only Expectations with an exact call count **can** have sequences" to better reflect that only expectations with exact call counts are allowed to be placed in a sequence.


asomers (approved):
Makes sense to me.

PR #672, opened 2025-11-11
Qualify send sync markers

rakai93:
When there is a struct in scope that is named `Send` or `Sync`, the code generated by `#[mockall::automock]` does not compile because it prefers the local scope over prelude traits. This PR fixes this issue by fully qualifying `::std::marker::Send` and `::std::marker::Sync` in the code generated by mockall.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   for Rust functions.
   ([#673](https://github.com/asomers/mockall/pull/673))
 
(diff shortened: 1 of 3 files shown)


asomers (changes requested):
Lol, that's a silly problem.  But your solution is a good one.  However, could you please add two things?

1. A CHANGELOG entry
2. A test case in mockall/tests

asomers:
If you rebase now, the test should be fixed.

asomers:
Mockall 0.14.0 is published, @rakai93

Issue #283, opened 2021-06-04
Add ability to reset / clear existing expectations?


Sushisource:
Hello, thanks so much for this library, super useful.

I have test setup code that will set a bunch of default expectations on mocks, but then some tests would like to alter the default expectations after the mock is built. Right now, if I just make another `expect_foo` call, that'll create an *additional* expectation, where I'd like it to actually replace the existing one. A nice way to do this might be something like a `remove_expectations_foo` method. Would this be a reasonable feature to add?

Thanks!

asomers:
Note that expectations are evaluated in LIFO order.  So if you set an expectation in your setup code and another one in a test case, the latter will be evaluated first.  That means the former doesn't usually cause any problems, unless there is a `.times` on it or something.  If you do have problems, I would suggest refactoring your setup strategy so as not to create expectations that you don't expect to fulfill.

Sushisource:
@asomers Hmm, the docs say the opposite about ordering: https://docs.rs/mockall/0.9.1/mockall/index.html#matching-multiple-calls ? It appears to be FIFO behavior to me.

And my use case is indeed using `.times`. I could change the setup strategy, but it'd be quite a bit more complicated than being able to wipe an existing expectation. Is your concern about having too many methods on the generated mock?

asomers:
Sorry; yes you're right it uses FIFO order.  Have you considered just removing `.times` from your setup?  In my own projects, the only methods I expect from the setup code are boilerplate stuff that could be called any number of times with the same result.  Any method that must have a fixed number of calls I expect from the test method.

Sushisource:
It's important to me to verify the number of calls is correct. I happen to have some pretty complex logic involved in setting up the expectations, which is why it's useful to have this common setup that works the majority of the time, but occasionally a test has reason to want to overwrite one of the method's expectations.

So, certainly possible to restructure how they're set up to accommodate that, but I figure some functionality like this is probably a useful tool to have in any case if it's not too much of an imposition on the library.

TomPridham:
this is also something that would be useful for us. we're mocking some static methods, which have global expectations. everything works great if all the tests pass. but, one test failing can bleed into another if the test fails prior to all the expectations being hit.

MikeColeGuru:
I have the issue of needing to change the return in the middle of a test.  Also, it would be great to be able to set a catch all.

[...]

PR #676, opened 2025-12-08
Provide clear_* and clear_and_expect_* mock methods to enable clearing the expectations array for automock

nobodie:
Partially addresses issue #283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+- Provide expectations clearing for automock.
+  Ex: Useful to configure a general mock behavior through a helper function,
+  and then specialize a single method by clearing it's initial behavior.
+  Partially addresses issue [#283](https://github.com/asomers/mockall/issues/283)
+
 ## [ 0.14.0 ] - 2025-11-22
 
 ### Added
--- a/mockall/tests/automock_struct.rs
+++ b/mockall/tests/automock_struct.rs
@@ -2,6 +2,7 @@
 //! automocking a struct
 #![deny(warnings)]
 
+use std::panic;
 use mockall::*;
 
 pub struct SimpleStruct {}
@@ -19,4 +20,48 @@ fn returning() {
     mock.expect_foo()
         .returning(|x| i64::from(x) + 1);
     assert_eq!(5, mock.foo(4));
+}
+
+#[test]
+fn clear() {
+    let mut mock = MockSimpleStruct::new();
+    mock.expect_foo()
+        .returning(|x| i64::from(x) + 1);
+    assert_eq!(5, mock.foo(4));
+
+    mock.expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(5, mock.foo(4));
+
+    mock.clear_foo();
+
+    mock.expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(6, mock.foo(4));
+}
+
+#[test]
+fn clear_and_expect() {
+    let mut mock = MockSimpleStruct::new();
+    mock.expect_foo()
+        .returning(|x| i64::from(x) + 1);
+    assert_eq!(5, mock.foo(4));
+
+    mock.expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(5, mock.foo(4));
+
+    mock.clear_and_expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(6, mock.foo(4));
+}
+
+#[test]
+fn calling_foo_without_expectation_after_clear() {
+    let mut mock = MockSimpleStruct::new();
+    mock.expect_foo()
+        .returning(|x| i64::from(x) + 1);
+    assert_eq!(5, mock.foo(4));
+
+    mock.clear_foo();
+    let result = panic::catch_unwind(|| {
+        mock.foo(4)
+    });
+    assert!(result.is_err());
 }
--- a/mockall/tests/automock_trait.rs
+++ b/mockall/tests/automock_trait.rs
@@ -2,6 +2,7 @@
 //! automocking a trait
 #![deny(warnings)]
 
+use std::panic;
 use mockall::*;
 
 #[automock]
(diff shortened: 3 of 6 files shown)


tgsmith61591:
@asomers any chance you've gotten to review this? This would be a really useful addition to the library.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+- Provide expectations clearing for automock.
+  Ex: Useful to configure a general mock behavior through a helper function,
+  and then specialize a single method by clearing it's initial behavior.
+  Partially addresses issue [#283](https://github.com/asomers/mockall/issues/283)
+
```
asomers, line 10:
```suggestion
## [ Unreleased ] - ReleaseDate

### Added

- Provide expectations clearing for a mock object's expectations.  Useful to configure general mock behavior through a helper function, and then specialize a single method by clearing it's initial behavior.
 ([#676](https://github.com/asomers/mockall/pull/676))
 
```

mockall/tests/automock_trait.rs
```diff
@@ -16,3 +17,48 @@ fn returning() {
...
+        .returning(|x| x + 1);
+    assert_eq!(5, mock.foo(4));
+
+    mock.clear_foo();
+    let result = panic::catch_unwind(|| {
+        mock.foo(4)
+    });
+    assert!(result.is_err());
+}
+
```
asomers, line 64: Rust already provides us with a great way to assert that something panics.  Let's use it.
```suggestion
#[test]
#[should_panic(expected = "@nobodie should fill in this string")]
fn calling_foo_without_expectation_after_clear() {
    let mut mock = MockSimpleTrait::new();
    mock.expect_foo()
        .returning(|x| x + 1);
    assert_eq!(5, mock.foo(4));

    mock.clear_foo();
    mock.foo(4)
}
```

mockall/tests/automock_struct.rs
```diff
@@ -20,3 +21,47 @@ fn returning() {
...
+    assert_eq!(5, mock.foo(4));
+
+    mock.expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(5, mock.foo(4));
+
+    mock.clear_and_expect_foo().returning(|x| i64::from(x) + 2);
+    assert_eq!(6, mock.foo(4));
+}
+
+#[test]
```
asomers, line 55: Just like in automock_trait, you should use `#[should_panic]` here.

mockall_derive/src/mock_function.rs
```diff
@@ -722,6 +722,119 @@ impl MockFunction {
...
 
+    /// Generate code for the clear_ method
+    ///
+    /// # Arguments
+    ///
+    /// * `modname`:    Name of the parent struct's private module
+    // Supplying modname is an unfortunately hack.  Ideally MockFunction
+    // wouldn't need to know that.
+    pub fn clear(&self, modname: &Ident)
+                  -> impl ToTokens
```
asomers, line 733:
```suggestion
    pub fn clear(&self, modname: &Ident) -> impl ToTokens
```

mockall_derive/src/mock_function.rs
```diff
@@ -722,6 +722,119 @@ impl MockFunction {
...
+        #[cfg(feature = "nightly_derive")]
+        let must_use = quote!();
+
+        let substruct_obj = if let Some(trait_) = &self.trait_ {
+            let ident = format_ident!("{trait_}_expectations");
+            quote!(#ident.)
+        } else {
+            quote!()
+        };
+        let docstr = format!("Clear the [`Expectation`]({}/{}/struct.Expectation.html) array, before creating one for mocking the `{}` method",
```
asomers, line 822:
```suggestion
        let docstr = format!("Clear the existing [`Expectation`]({}/{}/struct.Expectation.html) array, and add a new expectation for mocking the `{}` method",
```

mockall_derive/src/mock_function.rs
```diff
@@ -722,6 +722,119 @@ impl MockFunction {
...
+    // Supplying modname is an unfortunately hack.  Ideally MockFunction
+    // wouldn't need to know that.
+    pub fn clear_and_expect(&self, modname: &Ident, self_args: Option<&PathArguments>)
+                  -> impl ToTokens
+    {
+        let attrs = AttrFormatter::new(&self.attrs)
+            .doc(false)
+            .format();
+        let name = self.name();
+        let clear_and_expect_intent = format_ident!("clear_and_expect_{}", name);
```
asomers, line 791:
```suggestion
        let clear_and_expect_ident = format_ident!("clear_and_expect_{}", name);
```

mockall_derive/src/mock_function.rs
```diff
@@ -722,6 +722,119 @@ impl MockFunction {
...
+        } else {
+            quote!()
+        };
+        let docstr = format!("Clear the [`Expectation`]({}/{}/struct.Expectation.html) array, before creating one for mocking the `{}` method",
+                             modname, self.inner_mod_ident(), funcname);
+        quote!(
+            #must_use
+            #[doc = #docstr]
+            #(#attrs)*
+            #vis fn #clear_and_expect_intent #ig(&mut self)
```
asomers, line 828:
```suggestion
            #vis fn #clear_and_expect_ident #ig(&mut self)
```

mockall_derive/src/mock_function.rs
```diff
@@ -2576,6 +2694,11 @@ impl ToTokens for StaticGenericExpectations<'_> {
                         .unwrap()
                         .expect()
                 }
+
+                #v fn clear #ig (&mut self)
```
asomers, line 2698: This looks like it will clear expectations for all combinations of generic arguments.  Is that intended?  I would think that it would be more useful to clear them for only one specific set of generic arguments.

mockall_derive/src/mock_function.rs
```diff
@@ -722,6 +722,119 @@ impl MockFunction {
...
     }
 
+    /// Generate code for the clear_ method
+    ///
+    /// # Arguments
+    ///
+    /// * `modname`:    Name of the parent struct's private module
+    // Supplying modname is an unfortunately hack.  Ideally MockFunction
+    // wouldn't need to know that.
+    pub fn clear(&self, modname: &Ident)
```
asomers, line 732: This method contains much duplicated code from the `expect` method just above, and from the clear_and_expect method just below.  That's a shame.  But all three methods are called from exactly the same places.  What about combining them into a single method that generates code for three mock methods?  That could save some SLOC.

mockall_derive/src/mock_trait.rs
```diff
@@ -155,6 +155,21 @@ impl MockTrait {
                     meth.expect(modname, Some(path_args))
                 }
             }).collect::<Vec<_>>();
+        let clears = self.methods.iter()
+            .filter(|meth| !meth.is_static())
+            .map(|meth|
+                    meth.clear(modname)
+            ).collect::<Vec<_>>();
```
asomers, line 162:
```suggestion
            .map(|meth| meth.clear(modname)).collect::<Vec<_>>();
```

asomers (changes requested):
Thanks for the PR.  This looks like a good feature, and I intend to merge it.  Sorry that I've been so tardy on reviewing it.

In addition to my inline comments, I have a few general ones:

- You should add a test for clearing the expectations on a static method
- You should add tests for clearing the expectations on a generic method.  Does it clear all expectations, or only for a specific combination of generic arguments?
- `clear_and_expect` needs tests for ref and refmut methods.  Unless, as I suggested inline, you simply delete it.
- You should add a test that clearing an expectation with a `.times()` requirement does not cause a subsequent checkpoint to fail, if the mocked method was never called.
- Please add documentation.  I think this feature deserves a section in the crate-level documentation in mockall/src/lib.rs, just below the section on Checkpoints.

And have a happy MLK day.

PR #682, opened 2026-02-14
feat: Add spy! macro and #[autospy] attribute

### LINKED ISSUE #567: Feature request: create a `spy!` macro to provide a transparent mock but that can be asserted upon.
Similar to Java, it would have been nice to have a "spy" object that can be asserted upon: call counts with specific kind (or any) arguments, capturing arguments, and so on. Currently, we only have the `mock!` macro but then we will have to provide our own implementation. It would have been nice to have a `spy!` macro which, for example, would provide a "spying" implementation of a trait, based on some already existing implementation, which we would later just use.

exoego:
Closes #567 

Adds
- `spy!` procedural macro (analogous to `mock!`)
- `#[autospy]` attribute macro (analogous to `#[automock]`)
- `real.into_spy()` helper which is a shorthand for `SpyFoo::new(real)`

Design decisions:
1. Spy reuses the existing mock pipeline: `MockItemStruct` / `MockFunction` / `MockTrait` with a `GenerateMode` flag. This keeps code duplication low but ties spy behavior to mock internals.
2. Trait spies are generic: `SpyFoo<T: MyTrait>` accepts any type that implements the trait. Struct spies use a concrete type directly because inherent methods have no trait to use as a bound.
4. `impl Trait` return-types (including plain `async fn`) are not supported. Because Mockall converts these to `Pin<Box<dyn Trait>>`, which causes a type mismatch when calling the real implementation.
    - `#[async_trait]` works because it does this conversion before mockall processes the trait. This is documented as a known limitation.
5. `calling_real()` is needed for verified delegation. Unlike Java (Mockito) / Ruby(RSpec) spy frameworks that record all calls and let you check them after the fact, Mockall requires expectations to be set up before the call. So, users must explicitly use `calling_real()` when they want to both delegate and verify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,41 @@ mod tests {
 }
 ```
 
+## Spying
+
+Mockall also supports spies.  A spy wraps a real implementation and
+delegates method calls to it by default, but allows you to override
+specific methods with expectations.  The easiest way is to use
+[`#[autospy]`](https://docs.rs/mockall/latest/mockall/attr.autospy.html).
+It can spy on most traits, or structs that only have a single `impl` block.
+For things it can't handle, there is [`spy!`](https://docs.rs/mockall/latest/mockall/macro.spy.html).
+
+```rust
+#[cfg(test)]
+use mockall::autospy;
+
+#[cfg_attr(test, autospy)]
+trait MyTrait {
+    fn foo(&self, x: u32) -> u32;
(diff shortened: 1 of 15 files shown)


exoego:
@asomers Hello. I've rebased this into the master. It's great if this get reviewed.

PR #691, opened 2026-09-08
Suppress clippy::struct_field_names on generated structs

### LINKED ISSUE #689: Generated struct can trigger clippy::struct_field_names
# Steps to Reproduce

* Have the [clippy::struct_field_names](https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#struct_field_names) lint turned on
* Mock a trait whose methods, if used as struct fields, would trigger that lint
  * Example trait: [embedded_hal_async::digital::Wait](https://docs.rs/embedded-hal-async/1.0.0/embedded_hal_async/digital/trait.Wait.html)

Example code with this trait:

```rust
mod example {
    use embedded_hal::digital::{ErrorKind, ErrorType};
    use embedded_hal_async::digital::Wait;
    use mockall::mock;

    mock! {
        Wait {}
        
        impl ErrorType for Wait {
            type Error = ErrorKind;
        }
        
        impl Wait for Wait {
            async fn wait_for_high(&mut self) -> Result<(), <Self as ErrorType>::Error>;
            async fn wait_for_low(&mut self) -> Result<(), <Self as ErrorType>::Error>;
            async fn wait_for_rising_edge(&mut self) -> Result<(), <Self as ErrorType>::Error>;
            async fn wait_for_falling_edge(&mut self) -> Result<(), <Self as ErrorType>::Error>;
            async fn wait_for_any_edge(&mut self) -> Result<(), <Self as ErrorType>::Error>;
        }
    }
}
```

# Observed Behavior

```
error: all fields have the same prefix: `wait_for`

...

help: remove the prefixes
help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#struct_field_names

```

Coming from this generated struct:

```rust
#[allow(non_camel_case_types)]
#[allow(non_snake_case)]
#[allow(missing_docs)]
struct MockWait_Wait {
    wait_for_high: __mock_MockWait_Wait::__wait_for_high::Expectations,
    wait_for_low: __mock_MockWait_Wait::__wait_for_low::Expectations,
    wait_for_rising_edge: __mock_MockWait_Wait::__wait_for_rising_edge::Expectations,
    wait_for_falling_edge: __mock_MockWait_Wait::__wait_for_falling_edge::Expectations,
    wait_for_any_edge: __mock_MockWait_Wait::__wait_for_any_edge::Expectations,
}
```

# Desired Behavior

This lint is not triggered, as the shape of this struct is out of the user's control.

# Workaround

Turn off `clippy::struct_field_names` at the module level (after some experimentation I did not find a way to apply attributes to this particular generated struct).

# Possible fix

Add `clippy::struct_field_names` [here](https://github.com/asomers/mockall/blob/fc2c853cf7491c49147170d9f442eeb8d7589c2e/mockall_derive/src/mock_item_struct.rs#L425-L432) (I think that's the right place).  If this seems reasonable I would be happy to open a pull request with this change (are there any other places it should also be added?).

Alternately I could see how accounting for every possible lint might end up being kind of a never ending whack-a-mole exercise.  Is there, or could there be a way to apply specific attributes (`allow`/`expect` in this case) to this particular generated struct?

blowfishpro:
# Overview

One of the internal generated structs for each mocked trait or struct includes fields with the same names as the item's methods.

If all of the item's methods share a prefix or suffix, and the [clippy::struct_field_names](https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names) lint (part of `clippy::pedantic`) is enabled, then it will trigger the lint on the generated struct, as it uses the method names as struct field names.

Fixes asomers/mockall#689

# Questions

I wasn't totally sure about the file name for the test.  It uses `automock` as that was the simplest but this scenario applies to `automock` and `mock`.  I'm happy to rename it if desired.


mockall/tests/automock_clippy_struct_field_names.rs
```diff
@@ -0,0 +1,23 @@
+#! vim: tw=80
+//! One of the generated structs would trigger clippy::struct_field_names
+//! https://github.com/asomers/mockall/issues/689
+#![deny(clippy::struct_field_names)]
+#![expect(dead_code)]
```
asomers: Instead of allowing the dead_code lint, it's usually better in Mockall's test suite to make types pub.  Here, if you make `ExampleTrait` and `ExampleStruct` pub, then I think the dead code warnings will go away.

asomers (changes requested):

mockall/tests/automock_clippy_struct_field_names.rs
blowfishpro: Makes sense, and done (I also needed to make the struct methods public).

asomers (approved):
LGTM, thanks for the fix.

PR #692, opened 2026-09-17
Fix `#[automock]` on methods whose closures use associated types

### LINKED ISSUE #662: Automock errors if trait method argument contains a closure which uses associated type as an argument
Consider the following trait:

```rust
#[mockall::automock(type Item = types::MockSearchItem;)]
pub trait FooSearch {
    type Item: types::SearchItem;

    // ...

    fn remove_if<P>(&mut self, predicate: P) -> Vec<Self::Item>
    where
        P: Fn(&Self::Item) -> bool;
}
```

The `automock` macro will produce the following error:

```
ambiguous associated type

use fully-qualified syntax: `<__remove_if::Expectation as IntoAsyncIterator>::Item`, `<__remove_if::Expectation as IntoIterator>::Item`rustc[E0223](https://doc.rust-lang.org/error-index.html#E0223)
lib.rs(33, 16): original diagnostic
```

However, using the fully qualified associated type as suggested, it returns a different error:

```rust
#[mockall::automock(type Item = types::MockSearchItem;)]
pub trait FooSearch {
    type Item: types::SearchItem;

    // ...

    fn remove_if<P>(&mut self, predicate: P) -> Vec<Self::Item>
    where
        P: Fn(&<Self as FooSearch>::Item) -> bool;
}
```

```
bound `__remove_if::Rfunc: FooSearch` is not satisfied
the trait `FooSearch` is implemented for `MockFooSearch`
```

It's unclear whether this can be worked around.

YuEfSaEDU:
`#[automock]` failed with "ambiguous associated type" (E0223) when a method's closure argument referenced an associated type of the mocked trait, e.g. `where F: Fn(&Self::Item) -> bool`. The `#[automock(type X = ...)]` substitution machinery only visited method arguments and return types, so the unqualified `Self::Item` leaked into the generated private expectation module, where `Self` refers to the generated Expectation types rather than the mock. The compiler's suggested fully-qualified syntax could not work either, since the Expectation types do not implement the mocked trait, leaving no viable workaround.

This change extends `Attrs::substitute_trait` and `Attrs::substitute_item_impl` to also substitute within method signatures' generic-parameter bounds and where clauses, with a bound walker that handles the parenthesized `Fn(&Self::Item) -> bool` form legal in bound position. Signatures without associated types are untouched, and the regression tests cover where-clause, inline, and fully-qualified bound forms on both the trait and impl-block `#[automock]` flows.

Fixes #662


mockall_derive/src/automock.rs
```diff
@@ -308,6 +362,57 @@ mod t {
...
+            quote!(type T = u32;),
+            quote!(fn foo<F: Fn(&Self::T) -> bool>(x: u32);),
+            format_ident!("Foo"),
+            quote!(fn foo<F: Fn(&u32) -> bool>(x: u32);));
+    }
+
+    #[test]
+    fn closure_bound_multiple_traits_in_scope() {
+        // The concrete type must be substituted even when several traits
+        // define an associated type with the same name
```
asomers: What other trait defines an associated type with the same name?  Isn't this the same as `closure_bound_where_clause` but with the extra `+ Send` bound?  I don't see how the comment relates.

asomers (changes requested):
Hey, this looks really nice?  Could I ask that you also add a CHANGELOG entry?

asomers (approved):
Ahh, that comment makes more sense now.  Please rebase and squash.

PR #693, opened 2026-09-22
Removed unnecessary build scripts

ranger-ross:
This PR is an attempt to remove an unnecessary build script.
This crate was identified by an automated scan posted in the build-script working group ([#t-cargo/build-script > build script and proc macro allow list @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/628857-t-cargo.2Fbuild-script/topic/build.20script.20and.20proc.20macro.20allow.20list/near/625462381)).
The motivation is to reduce the usage build scripts across the ecosystem (for various reasons, security, reproduceable builds, compile times, etc)

The [`unexpected-cfgs`](https://doc.rust-lang.org/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table) lint in Cargo was added in 1.80 which is above this crate's MSRV (1.77.0) but only warns and the build still works.
```
❯ cargo build
warning: /home/ross/projects/mockall/mockall_derive/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
warning: /home/ross/projects/mockall/mockall/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s

❯ echo $?
0
```


ranger-ross:
If you don't like the warnings and are willing to bump the MSRV, I'd be happy to do that in this PR :)

